PR #5, opened 2018-07-19
Add Minimal OSX CI Build

insertinterestingnamehere:


insertinterestingnamehere:
Actually it looks like the build isn't expected to work on OSX right now. We can add a CI build for OSX if/when we add support for it.

PR #8, opened 2018-08-30
Fix CMake Source for CI tests

insertinterestingnamehere:
(first try it both ways to see if either setup currently works)

The CI servers recently changed their list of approved PPA repos for their APT plugin, so we'll likely have to find a different way to get CMake 3 on their ubuntu trusty image.
I'm currently experimenting with that here.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -157,11 +157,11 @@ jobs:
           make -j4
   "Arch":
     docker:
-      - image: archlinux/base
+      - image: base/archlinux
     steps:
       - checkout
       - run: |
-          pacman -Syy
+          pacman -Sy
           pacman -q -S --noconfirm gcc make cmake boost
           mkdir build
           cd build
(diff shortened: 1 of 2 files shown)


insertinterestingnamehere:
Okay, it appears that installing cmake3 from the default package repos just works in spite of it not being explicitly listed as an approved package in the Travis CI package whitelist, so we'll go ahead with that.

insertinterestingnamehere:
This did unearth an issue with our arch linux build, though that appears to be a bug with their package management. I'll see if there's an easy fix.

insertinterestingnamehere:
Okay, the arch build issues come from issues with their package repos and do not have an obvious fix here. I'm temporarily disabling that build. It can be re-enabled once they fix their packages. In general, we should use custom docker containers with circleci to avoid issues like this.

insertinterestingnamehere:
Ah! It appears that using a different arch docker image actually fixes it. The other one appears to have gotten out of date. I'll squash the relevant commits and then this should be good to go.

PR #12, opened 2019-02-14
Fix CircleCI build failures

insertinterestingnamehere:
Testing to see if switching the libstdc++ ABI used fixes the latest build failures.


insertinterestingnamehere:
I think the problem actually stems from different compiler versions. In particluar, conda-forge packages are no longer built with devtoolset. This makes it so that the latest boost from conda-forge can't find all the symbols it needs when it is used by the devtoolset provided compiler in our build here. I'm closing this and will experiment with solutions in a new PR.

PR #15, opened 2019-03-21
Installation Fixes

insertinterestingnamehere:
This should make it so that the headers and library install fine.


insertinterestingnamehere:
There are a few unrelated failures in the CI tests going on. Those can be addressed separately.

PR #17, opened 2019-03-21
Further CMake Fixes.

insertinterestingnamehere:
Use standard variables for specifying INCLUDE/BIN/LIB directories.
Also properly configure exported targets and components.


insertinterestingnamehere:
Okay, this covers most of the current CMake installation issues and appears to be working now. I'll leave it up for a little bit in case anyone gets the time to review it.

One thing related to this: we're currently exporting the flags used to build Galois to libraries that use an installed version of it via CMake. That's not a good way to do it. There's some machinery in place for a config.h that can hold the necessary defines and already provides versioning macros. Other flags used when building Galois probably shouldn't be exported at all. Sorting that out will have to be a separate PR later down the road though since it's another issue in its own right and I don't want it to hold this PR up.

PR #20, opened 2019-03-29
Run tests in CI

insertinterestingnamehere:
This is to test out how long running the apps on minimal inputs takes in a CI build. Testing in progress.


insertinterestingnamehere:
Okay, the tests currently only run in the CircleCI builds and are allowed to fail since we're in the process of fixing some of them. I'm going to wait to add them to the Travis CI builds since some of them currently hang when run there.

PR #23, opened 2019-04-22
Stop copying operators passed to parallel loops

insertinterestingnamehere:
This change is experimental and a little rough around the edges for the moment. Putting it up for discussion/feedback for now.


insertinterestingnamehere:
I've polished this up quite a bit more. It should be good to go. It's a net gain for the API of all three parallel loop constructs and some quick benchmarks show no meaningful performance differences in any of the apps.

This is a pretty significant change though, so I'll leave this up for another day or two. Review/feedback would be great.

insertinterestingnamehere:
https://godbolt.org/z/SKddoo is a scaled down demo showing how this actually works. Most of the rest of what's in this PR boils down to shuttling around template parameters and function arguments.

insertinterestingnamehere:
Okay, merging this now.

PR #33, opened 2019-05-21
Download fix

insertinterestingnamehere:
Make the CMake stuff for downloading the new sample inputs actually work.


insertinterestingnamehere:
Doesn't change anything that's actually tested in CI, so merging immediately.
TODO: test this in CI.

PR #43, opened 2019-09-25
Fix Debian CI failure

insertinterestingnamehere:
Debugging in progress. Will merge when finished.


insertinterestingnamehere:
The connection with the ubuntu package servers is spotty this morning, causing the 16.04 and 18.04 failures that are showing up here, but this fixes the Debian failure now, so I'll go ahead and merge.

PR #48, opened 2020-01-28
fix capture of object in examples for GCC <= 6

ddn0:
Older versions of GCC fail with

  [...]ExampleWrappedWorklist.cpp:49:20: error: cannot call member function ‘void ExampleWrappedWorklist::enqueue(int)’ without object
             enqueue(item + item);

Also, fix up some signed/unsigned warnings along the way.


insertinterestingnamehere:
Awesome! Thanks for this! The centos 7 clang build is an occasional random failure that pops up and the clang 3.8 and 3.9 failures we can ignore. I'll remove those builds shortly since I'm not convinced anyone is currently using Galois with those compilers.

PR #49, opened 2020-01-28
use uint{32,64}_t, ul and size_t consistently

ddn0:
Although the final widths may be the same, on some platforms, size_t and
uint64_t are not interchangable because they have different nominal
types (e.g., unsigned long long versus unsigned long) and vice versa,
and this causes build failures on MacOS due to unmatched template
instantiations.


insertinterestingnamehere:
Interesting that this is an issue on OSX since they're LLP64. Thanks for the fix!

PR #57, opened 2020-01-28
Add clang 10 to CI

insertinterestingnamehere:
In anticipation of upcoming release. (Tests will probably fail until the binary packages are public).


insertinterestingnamehere:
The LLMV PPA has the LLVM 10/Ubuntu Xenial packages up as of today, so I'm rerunning the tests to see if this is good to go now. Presumably the LLVM 10 release is happening any day now.

PR #58, opened 2020-01-29
Rest of the MacOS fixes

ddn0:
I didn't realize that `-DENABLE_DIST_GALOIS=ON` was a standard option 🤷‍♂.

This should be the rest of the necessary changes for *building*. There needs to be some more work to fix running on MacOS as the system introspection interface is different (e.g., no procfs).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,15 @@ git:
 
 matrix:
   include:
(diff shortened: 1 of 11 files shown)


insertinterestingnamehere:
LGTM. Thanks!

PR #59, opened 2020-01-29
Fix compiler warnings

ddn0:
With #58, this series of commits gets me to warning-free for
```
cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_DIST_GALOIS=ON
```
on my machine at least.


insertinterestingnamehere:
Thanks for tracking the warnings down! The CI failure is an intermittent one we haven't fixed yet, so I'm rerunning the offending build and then merging.

PR #60, opened 2020-01-31
Set _GNU_SOURCE when checking for sched_setaffinity

ddn0:
This needs to be defined otherwise sched_setaffinity will never be
found.

Tested by:
- [x] spinning up a ubuntu container and running `cmake -DUSE_STRICT_CONFIG=on ..`
---

Looks like Arch has updated its boost package

  https://www.archlinux.org/packages/?sort=&q=boost&maintainer=&flagged=

(Jan 30), which I think is causing the build failure.


insertinterestingnamehere:
Change LGTM. Agreed about the build. Merging till that can be addressed.

PR #65, opened 2020-02-01
Disable longjmp abort by default

ddn0:
I think I tracked down test failures in gmetis and torus{-improved} to
using longjmp rather than standard C++ exceptions. Given that this led
to incorrect results and segfaults, it seems safer to default to the
correct solution and give users a chance to override it with do_all
or no_conflict().

# How was this tested?
- [x] run the failing torus test 10ish times before and after recompiling with longjmp disabled; observe failures then lack of failures

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,7 @@ jobs:
           apt-get -q install -y libboost-iostreams-dev libboost-serialization-dev libeigen3-dev cmake gcc g++ zlib1g-dev libmpich-dev mpich
           mkdir build
           cd build
-          cmake -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=Release -DNUM_TEST_THREADS=2 -DENABLE_DIST_GALOIS=ON ..
+          cmake -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=Release -DNUM_TEST_THREADS=2 -DENABLE_DIST_GALOIS=ON -DUSE_LONGJMP_ABORT=OFF ..
           make input
           make -j2
           make test || :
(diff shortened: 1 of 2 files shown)


insertinterestingnamehere:
Thanks for tracking that down. It's unfortunate that this is the cause since stack unwinding in threads has major performance consequences in some cases. My inclination is to make exceptions the default but make it something an individual app can opt out of for now until a better fix for the exception issue is found. We should get input from others on this too though. Pinging @yishanlu, @darthscsi, and @amberhassaan to see if they have opinions since I know they've looked at this before.

darthscsi:
The correct solution is to fix the stack unwinding in libstdc++.  There is
a reader/writer lock which protects ELF header walking from
dlopen/dlclose.  Although in the parallel execution time, all access only
take the read lock on exception, the cache-line bouncing becomes a
serialization.  The fix I use to have was to implement a per-cpu
reader-writer lock in the abi runtime so exceptions only touched per-cpu
data.

If you don't use the fixed runtime, you have to use setjmp/longjmp for
scaling, but at the cost of having to track a few more things in the
runtime and having non-standard behavior.

Andrew

amberhassaan:
I agree with Andrew. We lose scaling when exceptions are enabled, which forced us to revive longjmp/setjmp. 

Just curious, does anyone know if libc++ (llvm one) has the same problem?

darthscsi:
The problem is in the itanium abi support library, not directly in the c++
runtime.

It's a non-trivial fix which impacts the quite a few code paths.

insertinterestingnamehere:
Yah, I'm very much in favor of getting this fixed upstream if possible, but there's still the question of what to do right now since it's not clear how long getting fixes accepted in all the needed places would take (or if all the needed changes will be accepted at all).

Here's a proposal for what to do now: let's make the exceptions/sjlj distinction something that can be configured completely at app build time via a preprocessor flag.

All that really needs to be done there is making the thread_local at https://github.com/IntelligentSoftwareSystems/Galois/blob/8fa18faea4cd0df855ae5d78a6aa2501c4cca6cd/libgalois/src/Context.cpp#L29-L51 be compiled into libgalois regardless of the flag setting and then moving the signalFailSafe and signalConflict functions into headers and giving them inline visibility. Then we can opt in to sjlj aborts for apps where that actually matters for performance.

This will make it so that things "just work" by default as far as correctness is concerned and make it so that we can still get good performance in cases where sjlj aborts actually matter. Isolating the sjlj cases will hopefully also let us move toward documenting the limitations of the sjlj aborts and confirming that those specific apps are written in a style that allows sjlj aborts to still run correctly.

How does that sound?

ddn0:
> [... moving setjmp/longjmp to an application-level flag ...]
> How does that sound?

I agree with everyone else that the ideal state is to use C++ exceptions as designed. If that requires upstream changes to address scalability concerns, we should figure out what exactly that entails and perhaps in the interim provide some user instructions on how to patch their system libraries.

I think that letting applications control their abort mechanism is giving them a lot of rope (not great) and in an ideal state we would have fixed the problem upstream so we will have to undo these changes later.

My counterproposal :D 
- figure out how to sensibly distribute a patched abi runtime
- rescope this PR to just changing the build configuration for the travis and circle-ci scripts so we can at least unblock our CI.

PR #66, opened 2020-02-02
Fix build on Arch Linux

ddn0:
The newer boost libraries 1.72 on Arch Linux only provide the
multithreaded variants. This causes a CMake configuration failure
because we explicitly set(BOOST_USE_MULTITHREADED OFF).

The most forward looking fix would be to remove the multithreaded
constraint but I'm not sure of the performance implications thereof. For
the time being to unblock the CI, just pin the boost library to a
previous version that does provide non-multithreaded variants.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -152,7 +152,7 @@ jobs:
       - checkout
       - run: |
           pacman -Sy
-          pacman -q -S --noconfirm gcc make cmake boost eigen openmpi
+          pacman -q -S --noconfirm gcc make cmake boost1.69 eigen openmpi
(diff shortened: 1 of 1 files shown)


insertinterestingnamehere:
I read up on this some more. Apparently boost multithreaded is just to guarantee that distinct objects can be safely used in distinct threads. It's a very weak form of thread safety. I suspect we'd run into correctness issues *without* using the boost multithreaded libraries. See https://stackoverflow.com/a/20991533/1935144 for details. The right fix is to use the multithreaded libraries in all cases. I'll submit a PR to take care of that.

PR #73, opened 2020-02-07
Bump to C++17 standard

### LINKED ISSUE #69: Barriers test crashes with recent versions of clang

ddn0:
Recently there have been issues with aligned structures but without
aligned allocations. This can create segfaults as a compiler might
generate code assuming a certain alignment (e.g., vmovaps) but the
allocator does not preserve the same alignment [1].

This really seems to be an issue with ::operator new pre-C++17 and the
most stable fix is to move to a newer standard.

[1] https://stackoverflow.com/questions/52000359

Closes #69

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,6 +45,10 @@ jobs:
       - checkout
       - run: |
           apt-get -q update -y
+          apt-get -q install -y apt-transport-https ca-certificates gnupg software-properties-common wget
+          wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | apt-key add -
+          apt-add-repository 'deb https://apt.kitware.com/ubuntu/ xenial main'
+          apt-get -q update -y
           apt-get -q install -y libboost-all-dev libeigen3-dev cmake gcc g++ zlib1g-dev wget libopenmpi-dev openmpi-bin
           mkdir build
           cd build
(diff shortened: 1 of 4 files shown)


insertinterestingnamehere:
I'm very much in favor of this, but input from others would be greatly appreciated. For older CMake versions we could just manually add `-std=c++17` or `-std=c++1z` to the compiler arguments.

ddn0:
@insertinterestingnamehere who would be good to bring into the discussion?

If we go to C++17, I think bringing up CMake to be the first version that supports `set(CMAKE_CXX_STANDARD 17)` seems reasonable.

I believe CMake would add `-std=c++1z` for older compilers.

insertinterestingnamehere:
@amberhassaan may have an opinion.

insertinterestingnamehere:
The only snag with old CMake versions is that our lmod modules at UT for CMake only give us CMake 3.3 with our gcc 6 package. It's not hard to fix the modules instead. @roshandathathri this will affect you until we get a newer CMake with the older gcc. Any preference? I'm inclined to let it be broken for a few days till we can get a newer CMake package.

insertinterestingnamehere:
Clarification: I think we should merge this and then deal with the CMake module issue.

roshandathathri:
For our heterogeneous code, we need to support gcc and other compiler versions supported by recent CUDA releases. For CUDA 9/10, that is gcc 7.

insertinterestingnamehere:
> For our heterogeneous code, we need to support gcc and other compiler versions supported by recent CUDA releases. For CUDA 9/10, that is gcc 7.

We won't be dropping any compilers with this. Just CMake versions older than 3.8.

insertinterestingnamehere:
Superseded by https://github.com/IntelligentSoftwareSystems/Galois/pull/79.

PR #82, opened 2020-03-26
Clang format

ddn0:
clang-format used to run on the repo but it wasn't enforced. Since we have CI now, it should be easy to keep everything nice and clean going forward.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,21 @@ version: 2
 # during the actual CI testing is no longer necessary.
 
 jobs:
+  "clang-format":
+    docker:
+      - image: debian:10
+    steps:
+      - checkout
+      - run: |
+          apt-get -q update -y
+          apt-get -q install -y apt-transport-https ca-certificates gnupg software-properties-common wget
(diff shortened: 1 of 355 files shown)


insertinterestingnamehere:
Thus far I had avoided enforcing clang-format use since I'm not sure everyone knows how to run this locally yet. This is probably a good idea though. Does anyone else have any input?

insertinterestingnamehere:
Rephrasing my comment slightly, I'm +1 on this, but I'd like to get buy-in from the rest of the group to avoid generating a lot of needless work.

l-hoang:
+1

roshandathathri:
+1

ddn0:
It looks like there are slight differences in output between how clang-format-9 and clang-format-10. Do people have a preference as the one to enshrine in the CI (for now)?

ddn0:
I also have a Dockerfile I can publish if folks have trouble installing clang-format.

l-hoang:
clang-format is on our machines, so that's not a concern.

RE: clang version; no preference mostly because I don't know the differences.

ddn0:
The support for this will come in with the merge from KatanaGraph. And the formatting itself should happen after all directory reorganization happens.

Issue #81, opened 2020-03-26
Crashes in GMETIS and torus_improved when built with clang


insertinterestingnamehere:
This is a longstanding issue that we've spent some time debugging. I'm going to document what's known about it here so the discussion can continue. It'd be nice to get this fixed for the next release.

insertinterestingnamehere:
Last time around we decided to go with https://github.com/IntelligentSoftwareSystems/Galois/commit/22532fae05c1a6454691ca8e04691c24e6124529 which improved the situation in some cases. After running the corresponding tests on various machines though, I think this was really only a partial fix. Failures are still present.

insertinterestingnamehere:
In particular, the failures appear to be architecture dependent. GMETIS never crashes on our skylake machines (the 56-core machines) regardless of the optimization setting, but torus still fails. On the other hand, both apps consistently fail on our Westmere based machines (the older 40-core ones) regardless of the optimization setting.

insertinterestingnamehere:
Recently @ddn0 noted that switching to exception-based aborts fixes the crashes, leading to the discussion in https://github.com/IntelligentSoftwareSystems/Galois/issues/71 (and related issues) about moving away from longjmp aborts. As I mentioned there though, given that exceptions acquire locks, it's likely that the locking involved is masking some other underlying synchronization bug (either in our code or in the compiler).

insertinterestingnamehere:
Here's a more thorough review of the architecture dependence: What matters is the architecture the binary is compiled for, not the architecture the binary is running on (provided the running architecture is new enough to run the compiled code). When compiled on Westmere (one of our 40 core machines), GMETIS and torus-improved both fail. In that case all parallel runs of GMETIS and all runs of torus-improved including sequential fail. When compiled and run on skylake (one of our 56 core machines), only parallel torus-improved fails and even then, the failure is inconsistent. The optimization flags don't appear to matter with the latest clang, so I propose we remove the code setting -O2.

insertinterestingnamehere:
Some other things to note: the most consistent GMETIS failures on older architectures are with the first test on small input, not the second. The second test fails only with some higher core counts when compiled on older architectures. Again, the optimization flags appear to have no effect here.

insertinterestingnamehere:
One other weird note: the crashes are significantly reduced in CI with clang 10. Also, example-torus (not just example-torus-improved) and dalaunayrefinement have recent unexplained failures in CI.

ddn0:
[...]

PR #83, opened 2020-03-26
Re-enable O3 with clang.

insertinterestingnamehere:
Rolls back our old partial fix for https://github.com/IntelligentSoftwareSystems/Galois/issues/81 since it's not actually helping much.


insertinterestingnamehere:
Superseded by a similar change included in https://github.com/IntelligentSoftwareSystems/Galois/pull/88

Issue #45, opened 2019-11-11
Fix Config File Machinery


insertinterestingnamehere:
There's currently some old broken machinery for generating a config file with defines that are set at CMake configure time. That header is no longer included anywhere. It should be fixed and used to provide info about how Galois is configured. Specifically, it should be used to provide:

 - version information
 - the type of aborts used inside the lockable type in the Galois shared object.

Both of those things are currently set as compiler flags in our current CMakeLists, but we shouldn't expect downstream users of an installed library to be aware of the flags used to compile it. That information needs to be included in an installed config header so they can check things with `#ifdef` statements. This is especially important for our operator aborting code because the headers currently will assume exception based aborts by default even though the library builds with setjump/longjump based aborts by default.

insertinterestingnamehere:
The generated config header should also have compiler-specific macros that enable us to silence warnings in a targeted way. Currently we're doing this with compiler flags, but those options aren't automatically carried to end users.

insertinterestingnamehere:
With the warnings overhaul, we need to remember to check if the Intel diagnostics we set manually are even needed anymore now that we use the more stringent warning flags.

PR #88, opened 2020-04-02
Outstanding commits from KatanaGraph

ddn0:
Cherry-picks from github.com/katanagraph/katana,

Basically the result of:
```bash
git log --no-merges kg/master ^iss/master | ... | tac | xargs git cherry-pick
``` 

It would be nice to get these in before any big refactor as they add stricter compiler warnings and various bug fixes.


insertinterestingnamehere:
BTW, I just merged https://github.com/IntelligentSoftwareSystems/Galois/pull/85. Some of the changes there may help with at least some of the CMake version issues in CI.

insertinterestingnamehere:
Also, if it makes things easier, I'm fine with dropping the OpenSUSE, Xenial, and CentOS 7 gcc builds in CircleCI. CentOS 7 is going off of full updates soon anyway. Xenial is already covered extensively in the Travis CI builds (which aren't showing up as checks right now for some reason). OpenSUSE is one of the least used Linux flavors we're currently testing and as little else to differentiate it from the other builds we're already running.

ddn0:
> Also, if it makes things easier, I'm fine with dropping the OpenSUSE, Xenial, and CentOS 7 gcc builds in CircleCI. CentOS 7 is going off of full updates soon anyway. Xenial is already covered extensively in the Travis CI builds (which aren't showing up as checks right now for some reason). OpenSUSE is one of the least used Linux flavors we're currently testing and as little else to differentiate it from the other builds we're already running.

Cool. I'll update this PR to include those changes.

Regarding the build matrix, is it frequently the case that we are looking to test the distributions themselves or just (compiler, lib, toolchain) compatibility? If the latter, it might be easier to just pick one distribution and install the various versions directly.

insertinterestingnamehere:
Travis CI already is set up to test the various compiler versions. I set up the CircleCI builds to test that specific common combinations of packages work in commonly used development environments. There is definitely some overlap between the two though.

insertinterestingnamehere:
Note, the Travis CI builds aren't currently showing up as checks for each PR. I've been trying to fix whatever setting is causing this but have been thus far unsuccessful.

insertinterestingnamehere:
See https://travis-ci.org/github/IntelligentSoftwareSystems/Galois

.travis.yml
```diff
@@ -26,11 +26,14 @@ matrix:
...
             - sourceline: 'ppa:ubuntu-toolchain-r/test'
+            - sourceline: 'deb https://apt.kitware.com/ubuntu/ xenial main'
+              key_url: 'https://apt.kitware.com/keys/kitware-archive-latest.asc'
+            - sourceline: 'ppa:mhier/libboost-latest'
           packages:
             - gcc-7
             - g++-7
             - cmake
-            - libboost-all-dev
+            - libboost1.70-dev
```
insertinterestingnamehere, line 36: Just to clarify, is this to confirm that we're good with the latest boost or is this because we're now requiring a more recent version? Are there additional warnings we need to suppress with older boost versions?

.clang-tidy
```diff
@@ -0,0 +1,74 @@
+---
```
insertinterestingnamehere, line 1: TODO for me: check that the clang-tidy stuff works out of the box on our development machines.

.git-blame-ignore-revs
```diff
@@ -0,0 +1,18 @@
+# Bulk-change revisions to ignore in git blame
```
insertinterestingnamehere, line 1: Cool. I didn't even know this was possible. Thanks!

libdist/CMakeLists.txt
```diff
@@ -7,9 +7,20 @@ set(sources
         src/NetworkIOMPI.cpp
         src/NetworkLCI.cpp
 )
+
+add_library(galois_dist_async_obj OBJECT ${sources})
+target_include_directories(galois_dist_async_obj PUBLIC
```
insertinterestingnamehere, line 12: We should confirm that the includes for libgalois and libllvm are picked up automatically by CMake and not list them here.

libgalois/CMakeLists.txt
```diff
@@ -43,55 +43,54 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
...
+    if (USE_STRICT_CONFIG)
+      message(FATAL_ERROR "Need sched_setaffinity")
+    endif()
   endif()
-
   list(APPEND sources src/HWTopoLinux.cpp)
 endif()
 
-add_library(galois_shmem ${sources})
+add_library(galois_shmem_obj OBJECT ${sources})
```
insertinterestingnamehere, line 54: Why the distinction between galois_shmem and galois_shmem_obj?

lonestar/experimental/sweeps/CMakeLists.txt
```diff
@@ -1,6 +1,5 @@
 if(("${CMAKE_VERSION}" VERSION_GREATER 3.8.0) OR ("${CMAKE_VERSION}" VERSION_EQUAL 3.8.0))
-    app(sweeps sweeps.cpp EXP_OPT)
```
insertinterestingnamehere, line 2: Actually this whole if/else can go. I put that in because I was using C++17 in sweeps/triangular solve/ILU before we decided to require C++17 and a more recent CMake.

ddn0:
@insertinterestingnamehere the latest errors are due to new code not passing the more strict compiler checks. I can continue fixing the new code but it might be easier to merge the changes in this PR so that committers can start fixing their own code instead.

test/gcollections.cpp
```diff
@@ -79,7 +79,7 @@ struct Heap<C, true> {
 };
 
 template <typename C>
-void testBasic(std::string prefix, C&& collection, size_t N) {
+void testBasic(std::string prefix, C&& collection, int N) {
```
insertinterestingnamehere, line 82: Anything that ever represents a "collection size" should always be size_t, but I suspect this patch is intended to  make the test match what the data structure does to remove a warning, so we can switch everything over to size_t later.

.circleci/config.yml
```diff
@@ -81,7 +94,7 @@ jobs:
           source activate
           # conda-forge used to use the devtoolset compilers.
           # Use the latest boost build that's still ABI compatible there.
-          conda install -y -q -c conda-forge cmake boost-cpp=1.69.0=h3a22d5f_0 eigen mpich=3.2.1=h26a2512_4
+          conda install --update-specs -y -q -c conda-forge cmake>=3.13 boost-cpp=1.69.0=h3a22d5f_0 eigen mpich=3.2.1=h26a2512_4
```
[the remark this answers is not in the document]
ddn0: I will admit I am not a conda expert :p 

My working assumption was that it is better to have a flattened dependency structure with everything handled within the same environment (and package manager). But I see that it would be better to avoid having old (boost) packages hold back the rest of our dependencies. Though, I'm not too inclined to rearrange too much here since this pain is mostly limited to certain CI configurations at the moment. 

Tangentially, does it make sense to make an "official" conda package for dev dependencies and eventually for released Galois? That might be the more sustainable way to turn all this effort finding compatible packages into something exploitable by all.

.clang-tidy
ddn0: I have a grand plan of applying `clang-tidy -fix` to the codebase, probably after enforcing clang-format.

It is pretty slow though since, I think, it reanalyzes headers for each compilation unit (approximately 45minutes for the entire code base: https://github.com/KatanaGraph/katana/runs/566296025).

My IDE just picks up this configuration so I've been applying the linter as I go for new files.

.travis.yml
ddn0: This is actually due to an interaction between:
- older versions of boost
- older GCC versions
- -Wimplicit-fallthrough
- our vendored libllvm

Basically, I was using `BOOST_FALLTHROUGH` to silence the implicit fallthrough warnings in libllvm. That doesn't work under certain older configurations.

We can relax the libboost constraint once we remove libllvm for instance.

libdist/CMakeLists.txt
ddn0: Do you mind if we defer this particular change for later?

I was planning to do this when we/I refactor our cmake usage to be strictly target based. And fix our installation :D

insertinterestingnamehere:
Sure. Let's do that.

libgalois/CMakeLists.txt
ddn0: Not my addition, but libcusp uses this to, I presume, build a fully resolved static library (and not depend on galois_shmem). Object libraries are the only cmake-portable way to do this. My change was to optimize this by actually using the compiled object files for both galois_shmem and galois_shmem_obj, rather than redundantly compiling them separately.

I think going forward I'd love to remove these weird edge cases (e.g., avoid object libraries altogether or use object libraries everywhere and have a final "library packaging" phase).

lonestar/experimental/sweeps/CMakeLists.txt
ddn0: Side note: we should probably come up with a plan for the experimental code. I'll make an issue.

insertinterestingnamehere:
@roshandathathri @gurbinder533  are the supercomputers you run your distributed apps on okay with the CMake >= 3.13 requirement?

CMakeLists.txt
```diff
@@ -1,22 +1,16 @@
-cmake_minimum_required(VERSION 3.8)
-
-include(GNUInstallDirs)
+cmake_minimum_required(VERSION 3.13)
```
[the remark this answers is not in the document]
ddn0: I think it would be good to move to "modern CMake". There doesn't seem to be an exact version when it starts, but it is around 3.11+, 3.12+:

  https://cliutils.gitlab.io/modern-cmake/

It looks like 3.13 was chosen because it is the first to support `add_link_options`.

CMakeLists.txt
```diff
@@ -77,100 +74,132 @@ set(USE_EMBEDDED_REVISION ON CACHE BOOL "Embed revision numbers")
...
+add_compile_options("$<$<COMPILE_LANGUAGE:CXX>:-g;-Wall;-Wextra;-Werror>")
+
+# GCC
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7)
+    message(FATAL_ERROR "gcc must be version 7 or higher. Found ${CMAKE_CXX_COMPILER_VERSION}.")
   endif()
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -wd68 -wd981 -wd383 -wd869 -wd2196 -wd279 -wd2504 -wd2943 -wd32013 -wd3373")
+
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 7 AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 8)
```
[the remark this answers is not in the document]
ddn0: I'm a little torn about warning suppression in source code, especially since there are multiple compiler/versions. Ideally we should make our headers as clean as possible, IMHO, and leave inline suppression as a last resort.

.circleci/config.yml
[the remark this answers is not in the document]
insertinterestingnamehere: Yah, I've been mostly leaving the CentOS 7 build alone here since we arguably don't need to keep it around after Q4 2020.

Once we fix more of our packaging issues I'd actually like to get a Galois package put together (just for the shared memory runtime) and maybe even submit it to conda-forge. I've already done most of the work for the recipe elsewhere and it makes packaging Python things that use Galois significantly easier. It's probably best to just continue to pick up the existing packages from either the main channel or conda-forge as dependencies though since that's where we'd be getting them once we provide our own package. I'm open to suggestions about other kinds of packages we might want to provide, but I haven't used conan and don't know how to build apt or deb packages so I'll defer to others on that front for now.

CMakeLists.txt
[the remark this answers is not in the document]
insertinterestingnamehere: Yah, add_link_options is definitely something we'd prefer to have. I'm fine with this, but we'll need to confirm that it's not breaking anything for others.

libdist/CMakeLists.txt
insertinterestingnamehere: That's totally fine with me. Several of these comments are meant as "we should fix this", not "we should fix this right now."

libgalois/CMakeLists.txt
insertinterestingnamehere: Got it. Thanks for the clarification. My preference is generally for shared objects everywhere, but I'm sure we can figure something sane out here.

lonestar/experimental/sweeps/CMakeLists.txt
insertinterestingnamehere: Agreed. Thanks.

CMakeLists.txt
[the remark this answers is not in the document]
insertinterestingnamehere: I've always viewed header cleanliness for stuff like this as something that has to be managed through sane use of macros. My main goal in this case is to make it so that an end user can include whatever they want from our library and not get errors/warnings even if they choose to compile their own code with `-Wextra -Werror`. My rule of thumb is that any compiler-specific configurations needed by a package belong in some well-isolated headers for that package. Build flags aren't picked up automatically by people who use a library but compiler-specific config headers are.

.travis.yml
insertinterestingnamehere: Got it. Relaxing this constraint after removing the vendored LLVM code is a great plan. Thanks for figuring that out.

.clang-tidy
insertinterestingnamehere: Sounds great. Let's try to do this after more of the new code for this release gets merged/finalized.

CMakeLists.txt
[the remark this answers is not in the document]
insertinterestingnamehere: I'll tackle this as a part of https://github.com/IntelligentSoftwareSystems/Galois/issues/45.

.circleci/config.yml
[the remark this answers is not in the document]
ddn0: I think conda will be a priority, especially once we get our Python support in place.

insertinterestingnamehere:
Okay, I think the needed issues are open. Thanks for helping with that. Any comments on the power of two number of threads thing?

.clang-tidy
insertinterestingnamehere: We have a clang-tidy executable already built with our existing clang and the existing script works, so no further intervention from me is needed for packaging there.

CMakeLists.txt
```diff
@@ -77,100 +74,132 @@ set(USE_EMBEDDED_REVISION ON CACHE BOOL "Embed revision numbers")
...
-enable_testing()
+if(NOT NUM_TEST_THREADS)
+  cmake_host_system_information(RESULT NUM_TEST_THREADS QUERY NUMBER_OF_PHYSICAL_CORES)
+endif()
+if(NUM_TEST_THREADS LESS_EQUAL 0)
+  set(NUM_TEST_THREADS 1)
+endif()
 
-###### Configure compiler ######
+if(ENABLE_HETERO_GALOIS)
```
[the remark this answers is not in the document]
roshandathathri: I would prefer that they are distinct flags. Heterogeneous codes require distributed Galois currently.

CMakeLists.txt
```diff
@@ -319,69 +346,87 @@ add_definitions(-D__STDC_LIMIT_MACROS)
 add_definitions(-D__STDC_CONSTANT_MACROS)
 
 ###### Test Inputs ######
-if(GRAPH_LOCATION AND EXISTS "${GRAPH_LOCATION}")
-  MESSAGE(STATUS "USING graph location ${GRAPH_LOCATION}")
+
+if(GRAPH_LOCATION)
+  MESSAGE(STATUS "Using graph location ${GRAPH_LOCATION}")
   set(BASEINPUT "${GRAPH_LOCATION}")
-elseif(EXISTS /net/ohm/export/iss/inputs)
```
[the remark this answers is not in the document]
roshandathathri: Yes, I would like to keep that location as well.

CMakeLists.txt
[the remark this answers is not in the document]
insertinterestingnamehere: USE_ARCH is the option for configuring the -march=<thing> option to the CPU compiler. Why would we keep that distinct?

CMakeLists.txt
```diff
@@ -319,69 +346,86 @@ add_definitions(-D__STDC_LIMIT_MACROS)
...
-      add_test(test-${type}-${app}-${thr} ${app} ${ARGN} -t ${thr})
-    endforeach(part)
+  set(options NOT_QUICK)
+  set(one_value_args)
+  set(multi_value_args REQUIRES COMMAND_PREFIX)
+  cmake_parse_arguments(X "${options}" "${one_value_args}" "${multi_value_args}" ${ARGN})
+
+  set(threads)
+  set(thr 1)
+  while (${thr} LESS ${NUM_TEST_THREADS})
```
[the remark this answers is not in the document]
ddn0: I have been thinking of these run-tests as more like holistic spot checks (for functional correctness), so good coverage would be "one threads", "some threads", "lots of threads" (e.g,, black box versus white box testing).

(I'd actually be happy to move from powers-of-two to just 1, n/2, and n.)

If there have been cases of issues related to the treatment of packages, I can see some value of being more specific, but then I'd push for a good unit test, independent of any app, first.

In general, I'm always a little wary of being too machine specific here because it can become a source of confusion for developers and maintenance burden down the line.

If these are intended to be performance tests, then I certainly agree with you.

I don't mind going back to multiples of 8 but I would probably also separate the more exhaustive testing from the quick tests and have a plan to run more aggressive testing out-of-line from PR integration tests.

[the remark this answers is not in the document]
roshandathathri: Good point. Why not do it reverse? Keep diving from max by 2 until 1, but run from 1 till max in that list.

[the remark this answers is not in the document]
insertinterestingnamehere: We've looked at running more aggressive performance regression tests as well, but those plans have been sidelined for the time being. +1 for the "go down by powers of 2" idea.

roshandathathri:
I made a pass over all the changes. No comments on anything else.

> @roshandathathri @gurbinder533 are the supercomputers you run your distributed apps on okay with the CMake >= 3.13 requirement?

One of the supercomputers does not have cmake >= 3.12. We can request them to upgrade but I'm not sure they will.

CMakeLists.txt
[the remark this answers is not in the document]
roshandathathri: Because ENABLE_HETERO_GALOIS has nothing to do with -march. It enables the distributed GPU code using Gluon.

insertinterestingnamehere:
Idk how much of a pain it'll be in their restricted environment, but CMake is just a command line utility so it's not super hard to use a locally downloaded binary. Would that work for you?

ddn0:
> One of the supercomputers does not have cmake >= 3.12. We can request them to upgrade but I'm not sure they will.

FWIW, it is pretty easy to install cmake from its release tarball:

```bash
curl -fL https://kitware... | tar -xzv -f - -C install-location
```

CMakeLists.txt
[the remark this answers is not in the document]
insertinterestingnamehere: Got it. Then we agree. For some reason USE_ARCH isn't applied when ENABLE_HETERO_GALOIS is set. Thanks!

gurbinder533:
Local installation should work fine.

roshandathathri:
Yes, local installation would work.

roshandathathri:
All the clusters we use now have cmake >= 3.13. So issue resolved.

PR #102, opened 2020-04-07
Don't error on newly introduced warnings

insertinterestingnamehere:
This should give the app authors time to fix these issues while not leaving CI totally broken.


insertinterestingnamehere:
@ddn0 the fixes for some of these weren't obvious to me, so I'm proposing we give @Breakinbad time fix them.

ddn0:
I think I got them all: https://github.com/IntelligentSoftwareSystems/Galois/pull/103

insertinterestingnamehere:
Great, then let's just go with that. Thanks!

Breakinbad:
Oh man that was fast! I just started to fix :p

PR #103, opened 2020-04-07
Fix warnings in bipart app

ddn0:

PR #104, opened 2020-04-07
Feature/unvendor libllvm

### LINKED ISSUE #77: libllvm too old to support option categorization
The latest (or some earlier versions) llvm commandline parsing library supports [option categorization](https://llvm.org/docs/CommandLine.html#grouping-options-into-categories), which makes the help information (output of -help option) more systematic and pretty. For example, we can categorize the common options like -runs, -t, etc. into a group, and other application- or even algorithm-specific options like -edgeTileSize into other groups. Current libllvm is too old to support this.

ddn0:
- [x] add CI dependencies
- [x] blocked on https://github.com/IntelligentSoftwareSystems/Galois/pull/103 to fix the CI issues

Closes #77


ddn0:
@insertinterestingnamehere I couldn't think of a better solution to 60da0e78104bc049b84c7e70f08e58bf98f8def6 (libllvm being compiled with libstdc++) than just removing the libc++ environment but I'm not too familiar with the options here. Are there any alternatives I should try for?

insertinterestingnamehere:
That's unfortunate, but there's not an easy alternative. If the libllvm ABI is dependent on the standard library ABI then we would need a libc++-based build of libllvm to be able to build with libc++. That's possible in general but it's more of a hassle than I'm willing to take on to keep that build around. It was really just there so we can benefit from the different set of assertions and to help us avoid relying on libstdc++ implementation details. This will likely affect the Travis libc++ builds too.

.travis.yml
```diff
@@ -16,10 +15,15 @@ matrix:
       before_script:
         - export CC=clang
         - export CXX=clang++
-        - brew install openmpi
+        - brew install openmpi llvm
         - mkdir build
+<<<<<<< HEAD
```
insertinterestingnamehere, line 20: Missed merge here.

ddn0:
@insertinterestingnamehere The last mile was figuring out that cmake was picking up Travis's copy of clang-7 with its broken libllvm rather than `llvm-7-dev`

ddn0:
@insertinterestingnamehere ping

.circleci/config.yml
```diff
@@ -41,13 +51,22 @@ jobs:
           apt-add-repository 'deb https://apt.kitware.com/ubuntu/ bionic main'
           apt-get -q update -y
 
-          apt-get -q install -y libboost-all-dev libeigen3-dev cmake gcc g++ libopenmpi-dev openmpi-bin ssh
+          apt-get -q install -y \
```
insertinterestingnamehere, line 54: Line breaks for package installs here is a great idea. Thanks.

.travis.yml
```diff
@@ -262,8 +228,12 @@ before_script:
       export BUILD_TYPE=Release
     fi
   - mkdir build
-  # Use apt-installed cmake rather than travis-provided one (/usr/local/cmake-3.12.4/bin/cmake).
-  - /usr/bin/cmake -S . -B build -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DENABLE_DIST_GALOIS=ON || exit 1
+  # Use apt-installed llvm-7-dev rather than travis-provided one which is
```
insertinterestingnamehere, line 231: Good catch. I wasn't aware that Travis provided one by default.

CMakeLists.txt
```diff
@@ -338,6 +326,9 @@ endif()
 find_package(Boost 1.58.0 REQUIRED COMPONENTS serialization iostreams)
 include_directories(${Boost_INCLUDE_DIRS})
 
+find_package(LLVM REQUIRED CONFIG)
+target_include_directories(LLVMSupport INTERFACE ${LLVM_INCLUDE_DIRS})
```
insertinterestingnamehere, line 330: Weird that they don't do this for us already, but that's not really our problem to fix.

ddn0:
>  I still need to make sure our internal builds of LLVM/clang get picked up without extra configuration hassle.

Cool. I'll let you merge in then? Feel free to rework this PR if you need extra configuration hooks.

insertinterestingnamehere:
Yep. I'll merge it soon.

insertinterestingnamehere:
Worth noting: this only works with LLVM builds configured with the non-default `LLVM_ENABLE_RTTI=ON` since boost's serialization library uses typeid and linking against an out-of-the-box build of LLVM without -fno-rtti set for our code gives a bunch of linker errors. It appears most existing binary packages for LLVM already include the needed typeinfo in their builds so we hopefully will only run into this when dealing with LLVM binaries built from source. I'm rebuilding our internal llvm/clang right now to confirm that that fixes the issue.

.travis.yml
ddn0: I can add something in the comment here about the RTTI requirement?

This is actually the root reason why we can't use /usr/local/clang-7 on travis.

insertinterestingnamehere: IDK if we need to mention this in the build script. We should check that LLVM_ENABLE_RTTI is enabled after doing `find_package(llvm ...)` though. I have a patch to do this. I'll push it to this branch and you can confirm that it seems reasonable to you.

insertinterestingnamehere: Can't figure out how to push to this PR so I'll just submit a separate one.

insertinterestingnamehere: https://github.com/IntelligentSoftwareSystems/Galois/pull/122

ddn0: Hmm. I thought "allow edits from maintainers" would allow you to write to my branch. Oh well. I added comments to #122 and cherry-picked your changes here (with modifications).

If the only remaining issue is providing more visibility for RTTI errors. I can work on that here and ping you for another review when it is ready?

PR #105, opened 2020-04-08
Feature/quick tests

ddn0:
It will probably be important to enable some testing in advance of merging lots of new functionality.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,9 +105,12 @@ jobs:
           cmake -S . -B /tmp/build -DENABLE_DIST_GALOIS=ON -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
           make -C /tmp/build input
           cmake --build /tmp/build --parallel 2
+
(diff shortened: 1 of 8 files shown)


ddn0:
> The CircleCI tests are now showing no tests found with the new ctest invocation.

Interesting, I think there must be something odd going on with CircleCI since when I run the same docker commands there is never an issue with reading the directory. I'll investigate further.

> Also, to clarify: the NUM_TEST_THREADS option is there to limit the number of threads that get started up in CI since the CI servers often report having many more threads than are actually available to the running build. I don't see how the new code is handling that.

Certainly, but I would prefer to have CI tests that are robust and can run on a variety of environments. If we have trouble running these tests on an (over)-loaded machine I would suggest fixing the tests for everyone rather than diving into configuration incantations. IMHO.

ddn0:
@insertinterestingnamehere I added issues for all the failing tests and added their executions here. I think getting this framework in place will be great to start keeping track of these latent bugs now :D

PR #122, opened 2020-04-13
Unvendor LLVM

insertinterestingnamehere:
Small changes on top of https://github.com/IntelligentSoftwareSystems/Galois/pull/104.


insertinterestingnamehere:
Arch failures look unrelated.

CMakeLists.txt
```diff
@@ -327,6 +327,7 @@ find_package(Boost 1.58.0 REQUIRED COMPONENTS serialization iostreams)
 include_directories(${Boost_INCLUDE_DIRS})
 
 find_package(LLVM REQUIRED CONFIG)
+message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
```
ddn0, line 330: We don't print out the version for our other libraries.

I went back and forth about adding something in the CMakeLists.txt file myself. In the end, I found just looking at `CMakeCache.txt` gave me all the information I needed for debugging though in the case of `find_package(LLVM)` you would have the infer the version from the `LLVM_DIR`. Other config modules end up populating an internal cmake variable `FIND_PACKAGE_MESSAGE_DETAILS_XYZ:INTERNAL` with version information.

For end users, it would probably be more useful to just put a version range check explicitly as they are not likely to want to read all this cmake verbiage.

CMakeLists.txt
```diff
@@ -328,6 +328,12 @@ include_directories(${Boost_INCLUDE_DIRS})
 
 find_package(LLVM REQUIRED CONFIG)
 message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
+if(NOT DEFINED LLVM_ENABLE_RTTI)
+  MESSAGE(FATAL "Could not determine if LLVM has RTTI enabled.")
+endif()
+if(NOT ${LLVM_ENABLE_RTTI})
+  MESSAGE(FATAL "Galois requires a build of LLVM that includes RTTI. Most package managers do this already, but if you built LLVM from source you need to configure it with `-DLLVM_ENABLE_RTTI=ON`")
```
ddn0, line 335: I think it is `message(FATAL_ERROR)` https://cmake.org/cmake/help/latest/command/message.html

insertinterestingnamehere: Yep. You're totally right. Thanks for catching that. Updated.

CMakeLists.txt
insertinterestingnamehere: Good point. Will do.

ddn0:
Fix for the arch issue: https://github.com/IntelligentSoftwareSystems/Galois/pull/104/commits/472ef3f9ee37582c4b2a0169a6f1f259a63aa577

I'll leave it up to you to pick which PR to merge

insertinterestingnamehere:
Thanks. I just cherry-picked the fix for arch. I'll merge here once the builds confirm that we're good.

ddn0:
Closes #104

Issue #99, opened 2020-04-07
Reinstate Default Graph Test Location


insertinterestingnamehere:
The code for using our default inputs directory got pulled out as a part of the merge in https://github.com/IntelligentSoftwareSystems/Galois/pull/88. We need to revive that.

PR #123, opened 2020-04-14
Default Inputs Directory

insertinterestingnamehere:
Fixes https://github.com/IntelligentSoftwareSystems/Galois/issues/99


insertinterestingnamehere:
Test failures appear unrelated.

CMakeLists.txt
```diff
@@ -266,6 +266,9 @@ include(CheckEndian)
 if(GRAPH_LOCATION)
   MESSAGE(STATUS "Using graph location ${GRAPH_LOCATION}")
   set(BASEINPUT "${GRAPH_LOCATION}/small_inputs")
+elseif(EXISTS /net/ohm/export/iss/inputs)
```
ddn0, line 269: I don't mind this change right now though I don't know what should happen in a world where most of the contributors and/or users to the codebase are not at ISS.

## On the test failures

Can you file an issue for each of them and mark the failing tests as `NOT_QUICK` now? The more we get used to seeing green the more likely we will want to keep it green :D

insertinterestingnamehere: Yah, I'm pushing to get people to run the tests consistently right now so this is there to try to make it as simple as possible. It's good to avoid hard-coded paths, but this seems like a reasonable price to pay to help people test in the short term.

insertinterestingnamehere: One potentially better option: we could move to checking for an environment variable and then just set that environment environment in the iss_load_modules script.

PR #126, opened 2020-04-14
Isolate Warning Suppression

insertinterestingnamehere:
WIP.


insertinterestingnamehere:
I'm going to put this temporarily on hold until merging https://github.com/IntelligentSoftwareSystems/Galois/pull/131 is taken care of.

ddn0:
So far seems reasonable to me as long as we are happy with what it takes to satisfy `-Wunused-parameter`. I'm also happy with blanket ignoring this particular warning if we think the resulting readability is worse than the alternatives.

insertinterestingnamehere:
Given that one of the fixes here is actually needed for gcc 7, I'm going to merge this once the tests pass. I'll come back for the no-deprecated-copy and no-cast-function-type warnings.

One open topic that can be addressed in a later PR: should we move away from the `GALOIS_UNUSED` macro and just use `[[maybe_unused]]`? Provided that this actually works with gcc 7, that seems like a better option long-term.

insertinterestingnamehere:
One additional note here: I made an attempt at distinguishing between arguments that are unused because a function conforms to some interface used elsewhere and arguments that are unused by mistake. I cleared out several examples where the arguments legitimately had no reason to be there. There may be a few left that I've now silenced by mistake, but one of the main benefits of this PR is that we will no longer have erroneous unused function arguments in new code.

PR #131, opened 2020-04-16
Merge dist-dev

insertinterestingnamehere:
Creating a PR here to check for test failures before dropping such a large set of changes into master.


l-hoang:
Looks like there are quite a bit of errors here. Anything I can do to help resolve them?

insertinterestingnamehere:
It looks like many of the failures shown are coming from the fact that the dist tests are just not running right at all in CI. It'd be great if you could look at that. The gcc 9 and clang 8-10 failures (the builds that failed much more quickly) are some kind of upstream packaging issue so hopefully those will get fixed soon without our intervention.

Here's how to debug the setup in CI: Make sure you've logged into CircleCI with your GitHub account, then click on one of the failed builds there. Also make sure you have an ssh key associated with your GitHub account. Click on one of the filing builds and then towards the top right there will be a button that says "rerun workflow". Click the arrow next to it and then hit "rerun job with ssh". Give it a few minutes to run. Once it finishes it will wait for a while instead of exiting the build and there will be instructions at the bottom of the build output showing how to ssh into the failing build.

insertinterestingnamehere:
Okay, I think all the current failures are caused by two specific issues:

- Some distributed tests are hanging/timing out with higher thread counts. I htink they are running out of memory. Examples include https://circleci.com/gh/IntelligentSoftwareSystems/Galois/5914 and https://circleci.com/gh/IntelligentSoftwareSystems/Galois/5912. We can try decreasing the size of the input even further, but rmat15 is already relatively small. We could also try building Galois as a shared lib to save some space. We haven't confirmed that that even works when building dist-galois though. I think it makes more sense to set NUM_TEST_THREADS in the build again. @ddn0 any input here?
- The distributed tests all use the `--env` flag, but that's mpich specific, so the openmpi builds all fail. https://circleci.com/gh/IntelligentSoftwareSystems/Galois/5913 is an example. @l-hoang I'd appreciate it if you could take a look at these failures.

l-hoang:
Unfortunately, this entire week I'm going to be busy with paper deadlines, so I won't be able to do much until next week.

Re: test failures: that's been something Roshan has been bugging me to do for months (some of the tests are very long running), and I haven't gotten around to it yet. It's definitely something I need to do.

Re: the -env flag; that's news to me, but I can also look into it.

roshandathathri:
I made the change to use --env flag for distributed tests. I'll look into making library-agnostic.

insertinterestingnamehere:
@roshandathathri That'd be great, thanks.

insertinterestingnamehere:
With the hanging tests we could also mark them as NOT_QUICK so that they just don't run in CI. I worry that that will include a large portion of the dist tests, but this will hopefully be temporary until someone can fix the slow ones.

ddn0:
> I think it makes more sense to set NUM_TEST_THREADS in the build again. @ddn0 any input here?

The original idea was to (a) separate out quick tests (i.e., can run in parallel, run in a couple of seconds) from not so quick ones and (b) run those quick tests in-line for CI.

If we stick to that idea, the path forward would be to confirm that the current set of failures are due to resource exhaustion and then mark some subset of distributed tests as not quick. E.g., mark only configurations with 2 or fewer threads as quick.

Alternatively, we could aim for a slower test configuration or reinstate `NUM_TEST_THREADS`, though I think partitioning along quick and not-quick is a little more clear cut.

insertinterestingnamehere:
Yah, ultimately I think it will depend on how many tests we lose by just marking the failures here as `NOT_QUICK`. I'll just have to try it and see.

insertinterestingnamehere:
Failures are unrelated. @ddn0, do you want to merge this now or continue in https://github.com/IntelligentSoftwareSystems/Galois/pull/137?

ddn0:
> Failures are unrelated. @ddn0, do you want to merge this now or continue in #137?

Feel free to merge now. I just wanted to see if we could be more surgical with test disabling. I can follow up later.

insertinterestingnamehere:
Yah, I spent some time trying to be more precise yesterday and it was looking like it was going to be a hassle, but if you get further with it then great. If not, we can wait for the original authors to clean things up. Thanks!

Issue #120, opened 2020-04-11
Merge GSimpleReducible with GBigReducible


ddn0:
The difference between GSimpleReducible and GBigReducible doesn't seem to be that much. I don't think anyone would mind if everything followed the interface of GBigReducible, which allows in-place updates. If we really wanted it to be simpler for users we could provide a simple wrapper for any pure out-of-place reduction function into an in-place one, e.g.,

```
T operator()(T, T) => T& operator(T&, const T&)
```

Though I think it is pretty rare that anyone creates new reducible types, so providing easier ways to do it likely is an over-optimization.

Actually, looking at the uses of GBigReducible, it looks like it is only used in the tutorial examples. We could just remove it altogether if there isn't a concrete user.

Thoughts @amberhassaan @swarnendubiswas ?

PR #140, opened 2020-04-23
Simplify Reducible

ddn0:
Previously, there were two different classes for reductions. One for
types that are efficiently copyable (GSimpleReducible) and one for types
are are expensive to copy (GBigReducible). STL containers manage to deal
with inefficient-to-copy elements without resorting to completely
separate implementations by using move semantics and adding API
adjustments like emplace. This commit does the same.

Replace the two reducible classes with a single Reducible class.
Reducible tries to use move semantics throughout, which will naturally
fall-back to copying if there is no appropriate specialization.

Remove the subclasses of GBigReducible, which existed to handle updating
container types while avoiding creating large intermediate values.
Instead, rely on the getLocal (previously peek_local) method to access
the current per-thread value and require clients to manipulate the value
directly.

This change does lose the clean abstraction property of the previous
approach but having wrapper classes for each composition of container
types seems unwieldy and difficult for users to understand and extend.


ddn0:
Closes #120

libgalois/include/galois/Reduction.h
```diff
@@ -20,546 +20,217 @@
...
+ *   T operator()(T lhs, T rhs)
+ *
+ * If T is expensive to copy, a moving merge function is more appropriate:
+ *
+ *   T& operator()(T& lhs, T&& rhs)
+ *
+ * IdFunc returns the identity element, which is used to initialize and reset
+ * the per thread values.
+ *
+ * The MergeFunc and IdFunc should have following property:
```
insertinterestingnamehere, line 47: They also need to be copy constructable. Lambdas happen to get this for free.

ddn0:
Sorry for the delay, I was finishing up https://github.com/IntelligentSoftwareSystems/Galois/pull/151

> As a concrete example, std::min and std::max return lvalue references (unfortunately). What happens if someone provides those as the operator?

I believe that `merge` would invoke the copy constructor or copy assignment operator around here: https://github.com/IntelligentSoftwareSystems/Galois/pull/140/files#diff-a9132b41a16748b78048e764721795d0R68 (Unless I'm missing something?)

Using std::max/min with Reducible is a little involved due to selecting the correct overload and converting from a function pointer to a general callable (perhaps there is a better way?). I added an example to `libgalois/test/reduction.cpp` to illustrate the hoops. This is one reason why we have the specializations GAccumulator, GReduceMin, ...

Regarding the larger question of well-formedness for any MergeFunc/IdFunc, I think that's what C++20 concepts are supposed to address?

libgalois/include/galois/Reduction.h
ddn0: Updated function doc.

insertinterestingnamehere:
> > As a concrete example, std::min and std::max return lvalue references (unfortunately). What happens if someone provides those as the operator?
> 
> I believe that `merge` would invoke the copy constructor or copy assignment operator around here: https://github.com/IntelligentSoftwareSystems/Galois/pull/140/files#diff-a9132b41a16748b78048e764721795d0R68 (Unless I'm missing something?)
> 

Okay, I see it now. Thanks for pointing me there. The merge function is doing an in-place reduction into a thread-local. The thread-local would get overwritten by whatever value is referred to by the returned lvalue.

> Using std::max/min with Reducible is a little involved due to selecting the correct overload and converting from a function pointer to a general callable (perhaps there is a better way?). I added an example to `libgalois/test/reduction.cpp` to illustrate the hoops. This is one reason why we have the specializations GAccumulator, GReduceMin, ...

Good question. I'm not sure of a fix for this offhand. Maybe using `std::invoke` instead of `operator()`? I'll look at it locally and let you know if I have any ideas. This shouldn't block the merge though.

> Regarding the larger question of well-formedness for any MergeFunc/IdFunc, I think that's what C++20 concepts are supposed to address?

Yep. That's the benefit concepts are supposed to give us. `static_asserts` provide an alternative to an extent, but they're sometimes still a hassle.

PR #137, opened 2020-04-21
Clean up recent dist branch changes

ddn0:

PR #145, opened 2020-04-23
Build fixes for pangolin

ddn0:
Should be merged after #137 because it builds on top of it.

I deleted various lonestarmine variants like sgl_edge_induced.cpp since they don't appear to compile with the current pangolin library. Please let me know if that is appropriate.

At some point we should remove `test_data` in favor of the same inputs system all other applications use. I filed a ticket: https://github.com/IntelligentSoftwareSystems/Galois/issues/143

A longer term question I had outside the context of this PR and just fixing the build: is it possible to not have header files that define `main()`?


roshandathathri:
LGTM

@chenxuhao please check

lonestarmine/include/lonestarmine.h
```diff
@@ -3,7 +3,7 @@
 #include <sstream>
 #include <iostream>
 #include "llvm/Support/CommandLine.h"
-#ifndef ENABLE_GPU
+#ifndef __GALOIS_HET_CUDA__
```
insertinterestingnamehere, line 6: Tangential to this, but we generally should avoid underscore prefixes on preprocessor symbols since that class of names is supposed to be reserved for the compiler. https://stackoverflow.com/a/228797 has details.

lonestarmine/README.md
```diff
@@ -1,25 +1,28 @@
 Overview of Graph Pattern Mining (GPM) in Galois
 ================================================================================
 
-This directory contains benchmarks that run using the Pangolin library[1].
+This directory contains benchmarks that run using the Pangolin framework [1]
+for efficient and flexible graph mining.  It uses the bliss library [2][3] for
+graph isomorphism check.  The license for this library is in the bliss
+directory: note that **it does not use the same license as the rest of
```
insertinterestingnamehere, line 7: Thanks for noting this more prominently.

libpangolin/gpu/pangolin/types.cuh
```diff
@@ -10,7 +10,7 @@ typedef uint8_t history_type;
 typedef unsigned char SetType;
 typedef unsigned long long AccType;
 
-#define MAX_SIZE     5
+#define PANGOLIN_MAX_SIZE     5
```
insertinterestingnamehere, line 13: Good idea prefixing stuff like this. Thanks.

libpangolin/include/pangolin/BfsMining/edge_miner.h
```diff
@@ -1,10 +1,10 @@
 #ifndef EDGE_MINER_H
 #define EDGE_MINER_H
-#include "miner.h"
-#include "quick_pattern.h"
-#include "canonical_graph.h"
-#include "domain_support.h"
-#include "embedding_list.h"
+#include "pangolin/miner.h"
```
insertinterestingnamehere, line 3: Thanks for getting rid of more relative paths and non-specific header includes.

lonestarmine/include/lonestarmine.h
ddn0: Yeah totally. I'd prefer a different name for this macro myself.

insertinterestingnamehere:
Forgot to mention on this: ideally we won't have headers that define `main()`. @chenxuhao why is this happening in `pangolin/BfsMining/engine.h`? Is there another way to do it?

PR #147, opened 2020-04-24
Fix dist tests

l-hoang:
reducing amount of compute dist tests do; PR here to run CI tests


roshandathathri:
LGTM

l-hoang:
Actually, still some failures I need to resolve according to some of these CI test results.

Have we found a way around the -env flag yet Roshan? It's causing some configs to completely fail.

roshandathathri:
@ddn0  has pushed a fix for env flag (not sure if it has been merged to master yet).

insertinterestingnamehere:
Someone ran into this over at https://github.com/dmlc/cxxnet/issues/211#issuecomment-338378389. It looks like with openmpi the `-x` flag can be used instead of `-env`. Offhand I'm not sure about any formatting differences between the two.

ddn0:
Yeah the fix (just merged into master) is to not set anything at all: 97e9839ce68c673d50de163b19472c2bfcc79aee

Spooky

insertinterestingnamehere:
Okay, just noticed that Donald fixed this already. That looks good.

l-hoang:
Cool; will merge and apply some more fixes then and try CI again.

l-hoang:
The -env fix revealed another class of errors that I'll be looking into. Some single process runs are failing as well, so that's good (easier to debug).

insertinterestingnamehere:
Great, thanks for looking in to these failures!

l-hoang:
Did some more investigating.

Seems to me like the issue is that the machine runs out of memory to support the amount of processes spawning in parallel. Each process takes up 1GB (seems to be preserved by Galois; tried a shared memory app as well), so if you have multiple tests running in parallel + each one potentially using 2-3 processes, memory runs out and segfaults occur.

I tested some of this by ssh'ing into the machine and spawning runs with conditions. Not having enough memory can either crash or even hang.

Any idea where I can change the amount Galois preallocs when it initializes the system to verify this hypothesis? Not too familiar with the memory system.

insertinterestingnamehere:
I don't know what could be causing that off the top of my head. 1GB is a massive memory footprint though, so we should try to get that down.

l-hoang:
There are currently 2 issues:
1) Stats generation fails non-deterministically sometimes; this is exceedingly rare, but needs to be fixed. It's not just a dist-tests issue either. @roshandathathri is currently on that.

2) Figured out yesterday that ThreadPool is allocating a decent chunk of memory which is causing the memory issues. Currently looking into that now.

libgalois/src/PerThreadStorage.cpp
```diff
@@ -38,15 +38,21 @@ galois::substrate::PerBackend& galois::substrate::getPPSBackend() {
 }
 
 #define MORE_MEM_HACK
+#ifdef GALOIS_LOW_MEM_FOOTPRINT
+// used for CI testing and other situations where galois memory
+// footprint needs to be low
+const size_t allocSize = 1 * (2 << 20); // 2MB
```
ddn0, line 44: There should be a nicer way to configure this rather than requiring a complete rebuild of the library.

I don't know the history of `MORE_MEM_HACK` but maybe a simple solution would be to just return to that code, which just allocates 4MB at a time, and file a ticket to verify later that there aren't any important performance regressions.

l-hoang: That is also fine with me. Not sure if anyone knows where that MORE_MEM_HACK came from (I don't); some context would be appreciated from anyone who does know.

l-hoang:
Seems like problem (2) is fixed (the LOW_MEM hack fixed all the segfaults; keeping it or not we can discuss later, but at least we know the problem), but problem (1) with non-deterministic stats hanging is still a problem; some CI tests are hanging and fail because of that from what I'm seeing.

libgalois/src/PerThreadStorage.cpp
ddn0: Looking at the code, I would guess that it was added because certain applications exhausted the buffer used to store per-thread data.

Correcting myself above, it looks like this is only a correctness issue (not a performance one) and libgalois will die if allocSize is too small, so it should be pretty straightforward to confirm if any apps actually require the larger allocSize and then determine if it is a true capacity problem or if we have too much fragmentation or internal leakage.

l-hoang: Interesting; so it actually never allocates more than that.

PerThread vectors and stuff of the sort might depend on the graph being used (e.g., amount you push to a vector depends on your input), so it may not be completely straightforward, but more static uses of PPS like accumulators and such can be easily tested with CI.

l-hoang:
Other than Alpine, seems like the failing tests occur due to the hanging stats bug.

Pushed in a change to check if the segfault is occurring because of PTS + many processes running on the machine at once or if it's something else.

(also, we should have nullptr checks for malloc anyways :P)

libgalois/src/PerThreadStorage.cpp
ddn0: I don't think there is an allocator that draws from the PerThread/PerSocket allocation pool so any PerThread<vector> like things would either use malloc or runtime/Mem.h (if they happened to use `gstl::Vector`. The memory for vector metadata (size, pointer to array) would be drawn from the PerThread allocation pool but that should be pretty small relative to the dynamic array itself.

l-hoang:
Seems like the most of the only failing tests now are the hang; Roshan has a fix for that and I'll get that in for testing soon.

The Alpine segfault seems to be another issue from the PTS and hang though; will investigate.

libgalois/src/PerThreadStorage.cpp
l-hoang: I see.

I can remove both mem hacks and use the huge page per thread; I'll try that in the next commit that will include Roshan's hanging fix.

l-hoang:
Hang is fixed thanks to Roshan. Also removed the memory hack define so it should be using the regular page alloc function now.

1) CentOS-8-clang failing for some odd reason on single process runs. CentOS-8-gcc is fine.
2) Alpine still failing/segfaulting.

Will be checking Alpine next.

l-hoang:
For multihost runs on CentOS, doing this fixes it:
https://github.com/openucx/ucx/wiki/OpenMPI-and-OpenSHMEM-installation-with-UCX#running-open-mpi-with-ucx

Still unclear how to get it fixed on Alpine, but I think the issue is probably similar as both fail the moment MPI is initialized.

@insertinterestingnamehere any idea how do deal with this?

l-hoang:
Will wait for Ian to look into the weird malloc/MPI issue.

Note that CentOS-8-clang passing is a lie: if you check the logs in question there's quite a bit of failures, but the fix for that is mentioned in another comment.

insertinterestingnamehere:
The MPI crash is not likely to be something I can figure out quickly. If you're not seeing a solution, I propose we stop building/testing dist-galois in those two build configs for the time being so we can get this merged. Thoughts?

roshandathathri:
Two questions:

(1) Are all dist-tests passing in some configuration?

(2) In those two build configs in which dist-tests are failing, are all dist-tests failing?

If (1) is true, then I'm fine with merging this to master.

If (2) is true, then I'm fine with not running dist-tests in those configs. However, let's continue to check if all the dist apps build fine in those configs if that is feasible.

l-hoang:
@roshandathathri 
(1) Yes.
(2) Yes, except on Alpine where 1 process runs pass if I recall correctly.

I'm fine with merging this in since it seems the major issues were resolved.

lonestardist/CMakeLists.txt
```diff
@@ -99,15 +99,23 @@ function(add_test_dist_and_verify app input type part N np)
 
 
   # Don't run dist tests in CI since they are broken in too many cases right now.
```
roshandathathri, line 101: Please remove this comment as well.

l-hoang: Done locally; will push once everyone has done a review to avoid pushing small changes like this.

roshandathathri:
LGTM

l-hoang:
I think Donald put himself on as a reviewer, so will wait for that + for CI to check.

libgalois/src/PerThreadStorage.cpp
```diff
@@ -37,17 +37,27 @@ galois::substrate::PerBackend& galois::substrate::getPPSBackend() {
...
+  }
+  return toReturn;
+}
 #else
-const size_t allocSize = galois::runtime::MM::hugePageSize;
-inline void* alloc() { return galois::substrate::MM::pageAlloc(); }
+const size_t ptAllocSize = galois::substrate::allocSize();
+inline void* alloc() {
+  // alloc a single page, don't prefault
+  void* toReturn = galois::substrate::allocPages(1, false);
```
ddn0, line 54: why not prefault? seems better to me to fault in now during initialization rather than possibly contending for the vm lock during parallel execution.

libgalois/src/PerThreadStorage.cpp
```diff
@@ -37,17 +37,27 @@ galois::substrate::PerBackend& galois::substrate::getPPSBackend() {
...
 
-#define MORE_MEM_HACK
 #ifdef MORE_MEM_HACK
-const size_t allocSize =
+const size_t ptAllocSize =
     16 * (2 << 20); // galois::runtime::MM::hugePageSize * 16;
-inline void* alloc() { return malloc(allocSize); }
-
+inline void* alloc() {
+  void* toReturn = malloc(ptAllocSize);
```
ddn0, line 44: I'd delete this code altogether.

roshandathathri:
+1 I'm all in for cleaner history

l-hoang:
Will make the suggested changes.

If I understand what you're saying about the rebase, then I guess you want something like this:

1) take diff from this PR
2) apply to master
3) commit to master
4) merge back to dist-dev

This will make it so master's history is clean.

The only problem is I already merged these changes into dist-dev after yesterday's meeting about workflow, so once dist-dev gets merged into master I don't see much of a difference (I think?).

roshandathathri:
Hehe, yes. I think it's fine for now.

ddn0:
> The only problem is I already merged these changes into dist-dev after yesterday's meeting about workflow, so once dist-dev gets merged into master I don't see much of a difference (I think?).

If this is already merged to some "visible" branch, then yes, it is probably too much to fix this. The time to fix this was when this change was merged into dist-dev. Though if one follows the tree analogy, we really should be merging dist-dev (unless fix_dist_tests is dist-dev?).

l-hoang:
fix_dist_tests was forked from master unfortunately, so following the tree analogy it's actually master then dist-dev.

And yes, the cleanup should have been done when I merged to dist-dev yesterday. Haven't heard of rebase --interactive before; will have to try it out.

For now I'll just follow the tree I guess.

ddn0:
> Haven't heard of rebase --interactive before

it's :fire:

l-hoang:
Guess I'll have to try it. :P

l-hoang:
Once these CI tests pass will merge into master and percolate down into dist-dev.

llvm separation is done in dist-dev; pull request for that will be opening soon after this PR is merged.

PR #149, opened 2020-04-25
fixing dist tc compilation error due to moderngpu

### LINKED ISSUE #148: lonestardist/tc doesn't build

vishweshjatala:
Replaced the older moderngpu function calls with new ones. I have tested locally and verified the correctness.  

Closes #148

### DIFF
--- a/lonestardist/tc/tc_cuda.cu
+++ b/lonestardist/tc/tc_cuda.cu
@@ -10,8 +10,7 @@ void kernel_sizing(CSRGraph &, dim3 &, dim3 &);
 #include "tc_cuda.cuh"
 #include "moderngpu/kernel_segsort.hxx"
 #include <cuda_profiler_api.h>
-using namespace mgpu;
-standard_context_t context;
+mgpu::standard_context_t context;
 #define WARP_SIZE 32
 
 inline __device__ unsigned long intersect(CSRGraph graph, index_type u, index_type v) {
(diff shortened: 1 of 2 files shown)


ddn0:
Closes #148 

Can you also re-enable tc for the all target (in `lonestardist/CMakeLists.txt`)?

lonestardist/tc/tc_cuda.cu
```diff
@@ -6,13 +6,12 @@
...
-#include "kernels/reduce.cuh"
+#include "moderngpu/kernel_reduce.hxx"
 #include "tc_cuda.cuh"
-#include "kernels/segmentedsort.cuh"
-#include "moderngpu.cuh"
-#include "util/mgpucontext.h"
+#include "moderngpu/kernel_segsort.hxx"
 #include <cuda_profiler_api.h>
-mgpu::ContextPtr mgc;
+using namespace mgpu;
```
ddn0, line 13: I generally like explicit namespaces (i.e., avoiding `using namespace XYZ`) as it is easier to see where symbols come from.

In this case, we still explicitly refer to `cub` like in `cub::BlockReduce` below. Is there a reason to use the `mpgu` namespace but not `cub`?

insertinterestingnamehere:
@vishweshjatala can we get this merged soon?

lonestardist/tc/tc_cuda.cu
vishweshjatala: Sure, we can use the explicit namespace. I will fix the code.

vishweshjatala:
@insertinterestingnamehere , I fixed the issues. You can merge if you are fine.

insertinterestingnamehere:
LGTM. Will merge once the tests pass. Thanks!

PR #150, opened 2020-04-25
Move aigRewriting and SPRoute out of experimental

insertinterestingnamehere:
Both of these apps still have warnings needing to be fixed. @Michael-JY-He is taking care of the ones in SPRoute. @yishanlu, there are quite a few in the aigRewriting code, not all with obvious fixes. Do you mind taking a look at them once this is merged?


yishanlu:
I can fix the warnings in aigRewriting code. Most of them are comparison between signed & unsigned integers, mixed usage of logic & arithmetic operations, and defined but unused functions. There are some warnings about the Morph_SepInOut_Graph, and I will discuss with you how to fix that if that is still happening after the merge.

insertinterestingnamehere:
That'd be great. Thanks. I started going through some of the warnings and noticed some used uninitialized warnings that looked like real bugs so I figured I'd defer to someone familiar with the app.

insertinterestingnamehere:
@ddn0 My plan is to merge this and then give the app owners a chance clean up their own warnings. Is that okay with you?

ddn0:
> My plan is to merge this and then give the app owners a chance clean up their own warnings. Is that okay with you?

Seems reasonable to me to get the app owners to handle warnings. Though, I do think it would be good to circulate some common fixes in advance so we get some consistency.

Some things I did for bipart (7aa106dbab133025 57d987d1147b1fc614) that are probably common for other apps:
- prefer direct initialization of numeric types (`size_t{0}`) over casting (`size_t(0)`/`static_cast<size_t>(0)`) as the latter is silently truncates.
- size_t != uint64_t (MacOS)

I'm not sure if we've reached total satisfaction on how to handle unused functions in debug versus release builds. Wrapping code in `#ifndef NDEBUG` may be an instance of a cure being worse than the disease. My preference is to restructure code to depend less on `assert` if possible. Though, we could go the other way and start having some utility macros to allow assert functions to be expressed more cleanly.

PR #156, opened 2020-04-29
Code cleaning and porting ktruss + initializing empty edge-data array with unweighted graph

nicelhc13:
Code cleaning and porting ktruss + initializing empty edge-data array with unweighted graph.
Tests were done with several input weighted graphs/ unweighted graphs.
BSPCoreThenTruss algo was removed due to its incorrect algorithm.
I was supposed to separate the branch into two, but mistakenly used one. Sorry about that.

### DIFF
--- a/libgalois/include/galois/graphs/ReadGraph.h
+++ b/libgalois/include/galois/graphs/ReadGraph.h
@@ -41,27 +41,32 @@ void readGraph(GraphTy& graph, Args&&... args) {
 
 template <typename GraphTy>
 void readGraphDispatch(GraphTy& graph, read_default_graph_tag tag,
-                       const std::string& filename) {
+                       const std::string& filename,
+                       const bool readUnweighted = false) {
   FileGraph f;
   f.fromFileInterleaved<typename GraphTy::file_edge_data_type>(filename);
-  readGraphDispatch(graph, tag, f);
+  readGraphDispatch(graph, tag, f, readUnweighted);
 }
 
 template <typename GraphTy>
 struct ReadGraphConstructFrom {
   GraphTy& graph;
   FileGraph& f;
+  bool readUnweighted = false;
   ReadGraphConstructFrom(GraphTy& g, FileGraph& _f) : graph(g), f(_f) {}
+  ReadGraphConstructFrom(GraphTy& g, FileGraph& _f, bool _readUnweighted)
+                         : graph(g), f(_f), readUnweighted(_readUnweighted) {}
   void operator()(unsigned tid, unsigned total) {
-    graph.constructFrom(f, tid, total);
+    graph.constructFrom(f, tid, total, readUnweighted);
   }
 };
 
 template <typename GraphTy>
-void readGraphDispatch(GraphTy& graph, read_default_graph_tag, FileGraph& f) {
+void readGraphDispatch(GraphTy& graph, read_default_graph_tag,
+                       FileGraph& f, const bool readUnweighted = false) {
   graph.allocateFrom(f);
 
-  ReadGraphConstructFrom<GraphTy> reader(graph, f);
+  ReadGraphConstructFrom<GraphTy> reader(graph, f, readUnweighted);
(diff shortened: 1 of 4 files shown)


nicelhc13:
One thing that I want to ask is preallocation.
How much memory should I preallocate in this app?
When I roughly tested it with some small preallocation, this app was crashed.

l-hoang:
The app shouldn't crash even if you preallocate 0; that seems to be something else going on.

nicelhc13:
OK, then can you recommend the amount of the preallocation?

l-hoang:
See the other benchmarks; it should be a function on nodes/edges of the graph.

Anyways, the point of what I was saying is that there seems to be another bug as no prealloc shouldn't cause crashes; if you don't prealloc, it means you allocate memory during runtime if you run out.

nicelhc13:
Yes. that's what I understand, but something happened. I didn't see this case from the other apps, so I am suspecting loop options. I agree with the 'function', but my question was about application-dependent information. I was wondering if you know the proper function for k-truss. It would be similar with the amount of memory that it'll use. I will look into it myself then. Thank you.

nicelhc13:
I am done with Ktruss.
Below are what I did from the recent commits.

**1) I reverted BSPCoreThenTruss back.**

After discussion with Yi-shan, I found that BSPCoreThenTruss is correct.
The reason was because I used non-symmetric graph. This was because my test was seg-faulted with symmetric graphs of our dist-inputs/symmetric directory and was not sure if it was OK with non-symmetric graph. It was my fault. 

Those symmetric graphs are unweighted, and libgalois did not allocate an edge data array for the unweighted input graphs. However, the k-truss is required to use the edge data array and it caused the segmentation fault. From the commit 1a40238 to b4eedbb, I add the option which allocates edge destination array even with unweighted input graphs to the Galois. 

**2) SymmetricGraph option**
From the above issue, I made the Ktruss forced users to specify -symmetricGraph option.
Through this, users can eliminate their confusions from the input graph formats.
(It seems like we should put this option to other apps using symmetric graphs if there exist)

**3) Add memory preallocations.**

**4) Apply clang-format**

roshandathathri:
Good job, Hochan. 

I think it's a good idea to make users pass -symmetricGraph flag to make sure that they understand symmetric graph is required for those apps.

nicelhc13:
@insertinterestingnamehere  Hi, Ian.
Can you give me any advice on this CI test failure?
To my understanding, the test failed at SPRoute point, which I didn't touch it.
It happened to all the pull requests. Thank you!

insertinterestingnamehere:
> @insertinterestingnamehere Hi, Ian.
> Can you give me any advice on this CI test failure?
> To my understanding, the test failed at SPRoute point, which I didn't touch it.
> It happened to all the pull requests. Thank you!

The SPRoute failure is from other code. If that's the only failure, this should be good to go.

libgalois/include/galois/graphs/ReadGraph.h
```diff
@@ -41,27 +41,39 @@ void readGraph(GraphTy& graph, Args&&... args) {
...
 template <typename GraphTy>
 void readGraphDispatch(GraphTy& graph, read_default_graph_tag tag,
-                       const std::string& filename) {
+                       const std::string& filename,
+                       const bool readUnweighted = false) {
   FileGraph f;
-  f.fromFileInterleaved<typename GraphTy::file_edge_data_type>(filename);
-  readGraphDispatch(graph, tag, f);
+  if (readUnweighted) {
+    //! If user specifies that the input graph is unweighted,
```
l-hoang, line 48: Doxygen syntax not required here; only required for defining vars and functions, not lines.

nicelhc13: Thank you for your comment. I'll keep it in my mind. 
But one thing that I want to mention is I summarized the comments cleaning as 'doxygen style', but also wanted to make consistent.

PR #162, opened 2020-04-30
Apply clang-format, and doxygen comment style

nicelhc13:
Possible todo: we can add transposeGraph option like the symmetricGraph option.


ddn0:
not blocking, but I'd hold off if possible on applying clang-format on files piecemeal.

It is just as easy to apply it over the entire codebase and enforce it going forward, especially since we are going to have a flag day when @insertinterestingnamehere moves all the directories around.

insertinterestingnamehere:
Yep. I'm planning on rerunning clang-format on all the things as a part of the reorganization.

PR #164, opened 2020-05-01
Remove unused bipart_csr application

ddn0:
Users should use bipart instead.

bipart_csr has already been removed from lonestar/CMakeLists.txt


ddn0:
errors are unrelated

PR #165, opened 2020-05-01
Add sanitizers

### LINKED ISSUE #97: Sanitizer CMake Configurations
We currently have a CMake configuration to enable using the address sanitizer. It'd be nice to have configurations that enable the other sanitizers as well.

ddn0:
Closes #97 

@roshandathathri , @l-hoang I eyeballed the results and there are some memory leaks and undefined behavior in the lonestardist tests. ~They don't show up now because this PR is pre-reenabling the tests.~ (edit: dists tests are back but now I've disabled sanitize builds due to lots of warnings) You can see what happens yourself with `cmake -DGALOIS_USE_SANITIZER="Address;Undefined" -DENABLE_DIST_GALOIS=ON ... && ctest -L quick`


l-hoang:
Stumbled upon your comment by accident (didn't get the notif as you edited the @ in after the fact I think). Going to @roshandathathri  so he can see any updates here too.

I pulled your branch and merged it locally into a copy of master to check the memory leaks, and I'm getting a few cmake issues.

CMake Error at /org/centers/cdgc/sw/miniconda/envs/cmake-3.17/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:164 (message):
  Could NOT find MPI (missing: MPI_C_FOUND MPI_CXX_FOUND)

      Reason given by package: MPI component 'Fortran' was requested, but language Fortran is not enabled.

Call Stack (most recent call first):
  /org/centers/cdgc/sw/miniconda/envs/cmake-3.17/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:445 (_FPHSA_FAILURE_MESSAGE)
  /org/centers/cdgc/sw/miniconda/envs/cmake-3.17/share/cmake-3.17/Modules/FindMPI.cmake:1717 (find_package_handle_standard_args)
  CMakeLists.txt:283 (find_package)

CMake Error at /org/centers/cdgc/sw/miniconda/envs/cmake-3.17/share/cmake-3.17/Modules/TestBigEndian.cmake:50 (message):
  no suitable type found
Call Stack (most recent call first):
  cmake/Modules/CheckEndian.cmake:2 (TEST_BIG_ENDIAN)
  CMakeLists.txt:260 (include)

Might have to do with Cmake install locally.

ddn0:
> I pulled your branch and merged it locally into a copy of master to check the memory leaks, and I'm getting a few cmake issues.

Thanks for pointing this out! There were some clangisms that weren't sanitized for gcc in the config code I took from LLVM. It is now fixed.

Also, this PR is now rebased on top of re-enabling dist tests.

In my quick skim, some of the leaks are from allocations in OpenMPI itself. Though we should tease out whether that is from our misuse or internal to OpenMPI. Other allocations seem to come from our own code. It is probably best to sort out the issues in our code first.

If you do decide to go down the path of looking into OpenMPI more, I have some small build scripts that will propagate sanitization through OpenMPI, which should give you better stack traces. They build on the work in `scripts/docker` to build memory sanitized dependencies. I didn't add them here because to do `-fsanitize=memory` (`GALOIS_USE_SANITIZER=Memory`) (NB: different from `-fsanitize=address`), we'd also need to recompile glibc.

l-hoang:
@ddn0 Not sure if this is what you expected, but I ran the latest commit locally with test failures here:

`PerThreadStorage.h:129:30: runtime error: load of null pointer of type 'char *'`

l-hoang:
It seems to be occurring on shared memory as well (same error). I am probably doing something wrong since the CI tests seem to pass.

.circleci/config.yml
```diff
@@ -10,6 +10,59 @@ version: 2
...
 
 jobs:
+  "CheckFormat":
+    docker:
+      - image: ubuntu:bionic
+    steps:
+      - checkout
+      - run: |
+          apt-get -q update -y
+          apt-get -q install -y apt-transport-https ca-certificates gnupg software-properties-common wget
```
insertinterestingnamehere, line 20: Aside from wget and gnupg, why do we need these other packages?

.travis.yml
```diff
@@ -25,7 +25,7 @@ matrix:
         apt:
           sources:
             - sourceline: 'ppa:ubuntu-toolchain-r/test'
-            - sourceline: 'deb https://apt.kitware.com/ubuntu/ xenial main'
+            - sourceline: 'deb https://apt.kitware.com/ubuntu/ bionic main'
```
insertinterestingnamehere, line 28: Thanks for remembering this.

scripts/docker/msan/build-boost.sh
```diff
@@ -0,0 +1,26 @@
...
+  -isystem ${LLVM_INSTALL_PREFIX}/include/c++/v1 \
+  ${MSAN_LINKER_FLAGS} \
+  -fsanitize=memory\
+  -w"
+
+mkdir -p "${SOURCE_DIR}"
+curl -fL https://dl.bintray.com/boostorg/release/1.73.0/source/boost_1_73_0.tar.bz2 | tar -xjv -f - -C "${SOURCE_DIR}"
+cd "${SOURCE_DIR}/boost_1_73_0"
+./bootstrap.sh --with-toolset=clang --with-libraries="${BOOST_LIBRARIES}"
+./b2 cxxflags="${MSAN_FLAGS}" linkflags="${MSAN_LINKER_FLAGS}"
```
insertinterestingnamehere, line 25: We probably want `-threading=multi` set as a flag for b2 here.

.circleci/config.yml
ddn0: `software-properties-common` adds apt-add-repository and the others allow verifying https repos. This is just the general incantation for taking a minimal ubuntu image to a slightly more useful one. We could filter it down on a repo/ppa basis for this particular use, but since these images are for dev environments instead of being deployed to production, it seems better to err on the side of ease of use rather than image size or excess dependencies.

ddn0:
> It seems to be occurring on shared memory as well (same error). I am probably doing something wrong since the CI tests seem to pass.

I didn't dig too much on this but setting the build to `Debug` or using clang instead of gcc seems to make this issue go away. It seems to be some interesting interaction between `gcc -fsanitize=undefined` and `thread_local` variables because as you note, if ptsBase were null, lots of applications should crash. If you think supporting gcc for this use case is important, please file an issue and someone can take a deeper look.

l-hoang:
Should be fine switching to the debug build; don't think it's particularly important to have gcc support for this if flipping to debug works.

Will try and let you know.

.circleci/config.yml
insertinterestingnamehere: Got it. Thanks for the clarification.

insertinterestingnamehere:
Yah, I wasn't aware that these even worked with gcc. My expectation was that I just had to use clang.

.circleci/config.yml
ddn0: Another bit of context now that I remember :D : 
I'm trying to reduce the configuration drift between travis and circleci. Travis uses a xenial-based VM and so these additional packages in CircleCI bring its minimal docker xenial image more inline with the Travis Xenial VM. Folks, now, should be able to "transcribe" travis sources and packages to their corresponding CircleCI equivalents.

From my brief experience, the UX for CircleCI seems nicer than Travis (ssh to debug FTW), and that's why I put the new CI configurations in CircleCI.

l-hoang:
I think I'm reproducing the issues now. Getting things like the following:

Serialize.h:240:12 runtime error null pointer passed as arg 1 which is declared to never be null
UndefinedBehavior Sanitizer: ...

Seems like overall the issues are in
1) bc_mr for all configurations
2) multiprocess execution for all apps

Will be looking into those. Should that be in a different pull request?

insertinterestingnamehere:
Given that things appear to be passing in CI right now, my preference would be to do that in a separate PR to minimize the risk of conflicts during the reorganization tomorrow morning. @ddn0 this looks good. Is it ready?

ddn0:
Yeah; it's ready. I'll merge it once the tests pass.

Thanks @l-hoang for kicking the tires around! It is good to get some feedback (gcc vs clang) even on the first iteration.

And yes, it is probably good to file the follow on work as a separate issue and set of PRs.

ddn0:
Looks like the Travis bipart failure predates this PR. Here is a failure from https://github.com/IntelligentSoftwareSystems/Galois/commit/2a2e5656d739c6228c996ce2f952ec65216aa22f  
  https://travis-ci.org/github/IntelligentSoftwareSystems/Galois/jobs/683113844

I've filled #171 to investigate.

PR #176, opened 2020-05-05
Lonestar Directory Reorganization

insertinterestingnamehere:


ddn0:
FWIW, it might be easier on the history to squash all these commits so we can ignore them en-masse (git-commit-ignore). I don't know if having commits on a per-app basis is that much more informative than "Reorganize directories to follow application areas" (or some such thing).

insertinterestingnamehere:
Sure. I can squash things if that'd be preferred.

insertinterestingnamehere:
Okay, this is ready to go. I'll merge it once the tests pass.

insertinterestingnamehere:
Quick note: In this PR I've also gotten rid of a bunch of the experimental code. I left lonestar/experimental for now though since some people are still working on apps there and we need to give them time to migrate their development somewhere else.

PR #178, opened 2020-05-05
Additional SPRoute Fixes

insertinterestingnamehere:
This clears out several (but not all) of the memory leaks found by the address sanitizer. I'm disabling the corresponding test in CI again until we get this to pass with the sanitizers.

### DIFF
--- a/lonestar/eda/cpu/SPRoute/utility.h
+++ b/lonestar/eda/cpu/SPRoute/utility.h
@@ -1207,7 +1207,7 @@ void StNetOrder() {
 
       gridsX = treeedge->route.gridsX;
       gridsY = treeedge->route.gridsY;
-      for (i = 0; i <= treeedge->route.routelen; i++) {
+      for (i = 0; i < treeedge->route.routelen; i++) {
         if (gridsX[i] == gridsX[i + 1]) // a vertical edge
         {
           min_y = min(gridsY[i], gridsY[i + 1]);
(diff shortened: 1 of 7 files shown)


lonestar/eda/cpu/SPRoute/utility.h
```diff
@@ -1207,7 +1207,7 @@ void StNetOrder() {
 
       gridsX = treeedge->route.gridsX;
       gridsY = treeedge->route.gridsY;
-      for (i = 0; i <= treeedge->route.routelen; i++) {
+      for (i = 0; i < treeedge->route.routelen; i++) {
```
insertinterestingnamehere, line 1210: @Michael-JY-He I'm not entirely certain this fix is right. I'd appreciate your input here.

insertinterestingnamehere:
Issue for remaining fixes: https://github.com/IntelligentSoftwareSystems/Galois/issues/179.

lonestar/eda/cpu/SPRoute/utility.h
Michael-JY-He: I don't remember I've change this line. Are you going to change it?

The line without "=" is correct. This loop tries to compute something between route point, which means it needs to iterate routelen - 1 times.

insertinterestingnamehere: Yes, the changes shown here are changes that I'm proposing we make in master. It'll be in master once I merge. Since this version is correct I'll go ahead and merge this. This will help with some (but not all) of the issues mentioned in https://github.com/IntelligentSoftwareSystems/Galois/issues/179. Thanks!

insertinterestingnamehere: Okay, these changes are visible in the current master now.

Michael-JY-He: Oh I was wrong. I'm Sorry.

The routelen is already -1.  The line with "=" is correct.

insertinterestingnamehere: Okay, feel free to change it back in your next round of edits. With the "=" there though, it has an out-of-bounds memory access that gets picked up by the sanitizer, so something is wrong in that case.

insertinterestingnamehere: Removing the "=" sign was my best guess at what the fix was without having more context on what the app was doing.

PR #180, opened 2020-05-05
Enforce clang-format

ddn0:


ddn0:
> Looks great, although I'm surprised that any differences showed up with clang-format again. I ran it this morning on the files that now show up as changed again.

There are slight differences between clang-format versions. Perhaps you were using clang-format-9 vs clang-format-10? Anyway, with the process checked by CI, it should be pretty easy to align everyone on the same version :D

insertinterestingnamehere:
> > Looks great, although I'm surprised that any differences showed up with clang-format again. I ran it this morning on the files that now show up as changed again.
> 
> There are slight differences between clang-format versions. Perhaps you were using clang-format-9 vs clang-format-10? Anyway, with the process checked by CI, it should be pretty easy to align everyone on the same version :D

That's what's weird. I was using clang-format 10.

insertinterestingnamehere:
Maybe it's not always idempotent? Either way, I think we should try this out and if anyone runs into issues where they run clang-format 10 locally and it still fails in CI we can rethink our approach.

ddn0:
> That's what's weird. I was using clang-format 10.

Hmm... that's annoying. We might investigate containerizing clang-format... Another alternative is to see if there is a set of clang-format options that is broadly compatible between released binaries. The current delta seems to be in the formatting of comments; we might investigate ignoring comments altogether:
- ReflowComments
- PenaltyBreakComment
https://clang.llvm.org/docs/ClangFormatStyleOptions.html

It would certainly reduce developer friction if everyone can reproduce the CI on their own machines.

insertinterestingnamehere:
Yah, it's unfortunate. My hope is that issues like this will be relatively rare, so we can just deal with them as they crop up.

PR #181, opened 2020-05-05
Fix the deprecated-copy warnings.

insertinterestingnamehere:


insertinterestingnamehere:
Okay, I just checked that this works with Boost 1.58 and gcc 8.1 or clang 10. I'm not sure what the old comment about boost counting iterator was referring to.

insertinterestingnamehere:
Test failures are unrelated. Merging.

PR #182, opened 2020-05-05
adding pta test

### LINKED ISSUE #175: points-to analysis test
Add a test for PTA.

l-hoang:
Resolves #175

### DIFF
--- a/libgalois/include/galois/runtime/Mem.h
+++ b/libgalois/include/galois/runtime/Mem.h
@@ -643,9 +643,23 @@ class FixedSizeHeap {
   SizedHeapFactory::SizedHeap* heap;
 
 public:
-  FixedSizeHeap(size_t size) { heap = SizedHeapFactory::getHeapForSize(size); }
+  FixedSizeHeap(size_t size) {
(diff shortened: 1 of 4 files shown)


l-hoang:
Looks like PTA has some memory issues that I'll have to look into.

l-hoang:
/root/project/libgalois/include/galois/runtime/Mem.h:650:45: runtime error: member call on null pointer of type 'galois::runtime::ThreadPrivateHeap<galois::runtime::FreeListHeap<galois::runtime::BumpHeap<galois::runtime::SystemHeap> > >'

This is the issue; will look into it.

l-hoang:
Not sure if that check I added for these failures was a sane check or if that heap for memory allocation being NULL is actually wrong check and it's supposed to be NULL, but it's causing a bunch of things to fail. Will check.

l-hoang:
This commit should fix it.

In addition to fixing the pta issue, added a few nullptr checks in Mem.h as part of debugging.

l-hoang:
Failures are CUDA tests; should be good to go after code review.

Issue #86, opened 2020-04-02
SPRoute


Michael-JY-He:
1. Clean up code. Remove warnings.
2. Prepare a small input for testing. Large input files.
3. Webpage.
4. Move to lonestar/EDA.

Michael-JY-He:
I thought I need to finish the webpage before closing lol?

insertinterestingnamehere:
Good point. Let's keep this open until the webpage is done too.

PR #187, opened 2020-05-07
Enforce no warnings in SPRoute.

insertinterestingnamehere:
@Michael-JY-He has already cleared away all the warnings in SPRoute, so this just re-enables `-Werror` for that app. After this, the only thing remaining for https://github.com/IntelligentSoftwareSystems/Galois/issues/86 is the webpage.


Michael-JY-He:
There are five failures but not from SPRoute.

PR #189, opened 2020-05-07
Re-enable bipart tests in CI.

insertinterestingnamehere:
WIP: Working to isolate the latest crash here.

### DIFF
--- a/lonestar/analytics/cpu/bipart/CMakeLists.txt
+++ b/lonestar/analytics/cpu/bipart/CMakeLists.txt
@@ -2,4 +2,4 @@ add_executable(bipart bipart.cpp Coarsening.cpp Metric.cpp Partitioning.cpp Refi
 add_dependencies(apps bipart)
 target_link_libraries(bipart PRIVATE Galois::shmem lonestar)
 install(TARGETS bipart DESTINATION "${CMAKE_INSTALL_BINDIR}" COMPONENT apps EXCLUDE_FROM_ALL)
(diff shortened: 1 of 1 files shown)


insertinterestingnamehere:
Okay, I can't reproduce the test failures here on an Ubuntu 18 VM, so I suspect that the errors are caused by resource limitations in CI. Still investigating though.

insertinterestingnamehere:
@Breakinbad I fixed this by switching back to the `ibm01.hgr` file. It was running out of memory with the other one. Is that alright with you?

insertinterestingnamehere:
@ddn0 I turned off parallel testing in the Travis CI tests to avoid running out of memory. I'll try re-enabling it, but we may need to make that change permanent. Is that alright with you?

insertinterestingnamehere:
Yah, I'll squash or reset appropriately depending on if it still works with parallel testing re-enabled.

insertinterestingnamehere:
I had started talks with CircleCI for one last year and the discussions stalled out. I can try again with them or Travis CI or both. It's a good idea. We can revisit that next meeting. Another option would be to set up configurations for Azure Pipelines and/or GitHub actions. I already have a semi-usable Azure Pipelines config elsewhere.

insertinterestingnamehere:
Okay, disabling parallel tests wasn't actually needed, so I've squashed away the commit then revert that was here before.

ddn0:
> GitHub actions

Example here with caching, etc.

https://github.com/KatanaGraph/katana/tree/master/.github/workflows

I'd suggest just sticking to one system whatever it is. CI services are pretty much a commodity at this scale, so I'd just pick whatever is easiest to use or cheapest.

Issue #92, opened 2020-04-07
USE_ARCH flag work for Heterogenous Galois too


ddn0:
Not a blocker, but we should make the USE_ARCH flag work everywhere. Even for Heterogeneous Galois, we can still set the architecture flag for all the CPU code generated.

_Originally posted by @insertinterestingnamehere in https://github.com/IntelligentSoftwareSystems/Galois/pull/88_

roshandathathri:
I did not understand this. Why?

ddn0:
It is a little unexpected that `USE_ARCH` is completely ignored if `ENABLE_HETERO_GALOIS=ON` since those two options are not related to each other for CPU code.

https://github.com/IntelligentSoftwareSystems/Galois/blob/master/CMakeLists.txt#L69

insertinterestingnamehere:
Yah, it's an unexpected behavior issue. I'm not aware of anything that makes this urgent. The current behavior is just weird.

roshandathathri:
What is USE_ARCH expected to do?

I will be renaming ENABLE_HETERO_GALOIS to GALOIS_ENABLE_GPU (and merging it with USE_GPU). That flag will enable single/distributed GPU code for lonestar (analytics/mining/scientific).

insertinterestingnamehere:
USE_ARCH sets the appropriate `-march=...` flag for the CPU code.

roshandathathri:
I don't see USE_ARCH being used anywhere currently. It was being used for KNL if I remember right. But where is that code in our cmake files?



> https://github.com/IntelligentSoftwareSystems/Galois/blob/master/CMakeLists.txt#L69

I will drop those lines. They are not related/required.

insertinterestingnamehere:
It's in https://github.com/IntelligentSoftwareSystems/Galois/blob/master/cmake/Modules/CheckArchFlags.cmake which defined variables that then get pulled in here: https://github.com/IntelligentSoftwareSystems/Galois/blob/master/CMakeLists.txt#L132.

insertinterestingnamehere:
My best guess is that the if statement at https://github.com/IntelligentSoftwareSystems/Galois/blob/master/CMakeLists.txt#L69 was added to work around some kind of compiler error for the heterogeneous case. @ddn0, you added that, do you remember why it's there?

roshandathathri:
[...]

PR #193, opened 2020-05-07
ci: Add GPU builds

### LINKED ISSUE #190: Add GPU builds in CI
We can't run them but we can at least check if the GPU build works

ddn0:
The CI won't pass until at least these issues are resolved:
- #195 
- #196 
- #194

(just for historical context, I noticed these errors while discussing #92 though they are not related to that issue itself)

Closes #190

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,10 +17,13 @@ jobs:
       - checkout
       - run: |
           apt-get -q update -y
-          apt-get -q install -y apt-transport-https ca-certificates gnupg software-properties-common wget
+          apt-get -q install -y apt-transport-https ca-certificates git gnupg software-properties-common wget
           wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key 2>/dev/null | apt-key add -
           apt-add-repository -y 'deb https://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main'
           apt-get -q update -y
+      - run: git submodule sync
+      - run: git submodule update --init
+      - run: |
           apt-get -q install -y \
             clang-format-10
 
@@ -34,7 +37,7 @@ jobs:
       - checkout
       - run: |
           apt-get -q update -y
-          apt-get -q install -y apt-transport-https ca-certificates gnupg software-properties-common wget
+          apt-get -q install -y apt-transport-https ca-certificates git gnupg software-properties-common wget
           wget -O - https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null | apt-key add -
           wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key 2>/dev/null | apt-key add -
           apt-add-repository -y 'ppa:ubuntu-toolchain-r/test'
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -134,6 +146,62 @@ jobs:
...
+            g++ \
+            gcc \
+            libboost-all-dev \
+            libeigen3-dev \
+            libopenmpi-dev \
+            llvm-7-dev \
+            openmpi-bin \
+            ssh
+
+          cmake -S . -B /tmp/build -DENABLE_DIST_GALOIS=ON -DENABLE_HETERO_GALOIS=ON
```
roshandathathri, line 174: -> cmake -S . -B /tmp/build -DGALOIS_ENABLE_DIST=ON -DGALOIS_ENABLE_GPU=ON

.circleci/config.yml
```diff
@@ -134,6 +146,62 @@ jobs:
...
```
roshandathathri, line 202: -> cmake -S . -B /tmp/build -DGALOIS_ENABLE_DIST=ON -DGALOIS_ENABLE_GPU=ON

.gitmodules
```diff
@@ -1,6 +1,3 @@
-[submodule "bliss"]
```
roshandathathri, line 1: Why is this removed? I thought this was required for pangolin.

ddn0: It is already vendored in external directly rather than as a submodule. The submodule itself was just something @chenxuhao made to track bliss upstream but it seems simpler to avoid inserting an intermediary.

.circleci/config.yml
ddn0: Thanks for reminding me! Though this change only needed if this PR goes in after #199 . Otherwise it is actually correct as it stands.

.gitmodules
roshandathathri: Yes, it's better it's not a submodule but that is not related to this PR. As long as pangolin builds don't break, it's fine to be included in this PR.

ddn0: FWIW, the reason the commit is here is because we have to enable submodules for GPU builds, so I took the opportunity to clean up our submodule configuration, which has these unused entries.

roshandathathri: Ok, makes sense.

roshandathathri:
LGTM

.travis.yml
```diff
@@ -2,11 +2,8 @@ dist: bionic
 
 language: c++
 
-# At the moment, our LCI submodule has an additional submodule within it
```
insertinterestingnamehere, line 5: Good catch. This comment is outdated now that the LCI submodule is gone.

ddn0:
> It'll be great to have these working. Thanks for putting this together. One thing wasn't obvious to me: why do we need to check out submodules in all the builds when the submodules are really only used for the GPU code?

You are correct, but I figured it would be easier for maintenance and reduce the number of moving parts if we just standardized what it means to check out the repo (checked out or not vs checked out with submodules, checked out without ....). Alternatively, I imagine my future self to be a little surprised to see CI fail if I added a submodule that the CPU code needed, and my future self would have to add this same submodule checkout code anyway.

Issue #153, opened 2020-04-27
Rename __GALOIS_HET_CUDA__


ddn0:
Tangential to this, but we generally should avoid underscore prefixes on preprocessor symbols since that class of names is supposed to be reserved for the compiler. https://stackoverflow.com/a/228797 has details.

@insertinterestingnamehere

https://github.com/IntelligentSoftwareSystems/Galois/pull/145

roshandathathri:
+1 

We should rename all ___X preprocessor symbols we use to GALOIS_X

roshandathathri:
Fixed in PR 199

Issue #132, opened 2020-04-17
AIG rewriting


yishanlu:
1. Move from lonestar/experimental/aigRewriting to lonestar/eda/aigRewriting
2. Sync with changes for `galois::graphs::First_SepInOut_Graph` as in #115.
3. Testcases
4. Webpage

yishanlu:
Thanks @insertinterestingnamehere for taking care of 1. & 2.

The warnings are fixed except for 40 `-Wunused-functions` and 1 `-Wunused-but-set-variable` (actually used in `assert()`).

We can use EPFL combinational circuits as test cases for sanity check. This is available on ohm, and the compiled executable runs normally with one of the circuits there. So what is required for test cases?

yishanlu:
Confirmed using ABC from UC Berkeley that mem_ctrl.aig from EPFL benchmark suite is equivalent to mem_ctrl_rewritten.aig, the output generated by aigRewriting. This circuit is available at `/net/ohm/export/iss/inputs/eda/logic-synthesis/EPFL/random_control/mem_ctrl/aiger/mem_ctrl.aig`.

Will add a test for AIG rewriting using this circuit and the generated output.

insertinterestingnamehere:
A test is the only thing still needed here.

yishanlu:
The following changes are available in out private master branch:
1. Converted the command line parsing to using `LonestarStart()` for uniform testing commands. 
2. Added a test (using `/net/ohm/export/iss/inputs/eda/logic-synthesis/EPFL/arithmetic/adder/aiger/adder.aig`).

In case you need the input path for test setup, the input AIGs are available at `/net/ohm/export/iss/inputs/eda/logic-synthesis/EPFL/*/*/aiger/*`.

roshandathathri:
Many CI builds are failing. Failures can be seen in this PR:
https://github.com/IntelligentSoftwareSystems/Galois/pull/199

insertinterestingnamehere:
I'll just disable that test in CI until we get it fixed.

yishanlu:
I saw #200 being merged. Are we good with AIG rewriting?

insertinterestingnamehere:
WRT the crashes not fixed in https://github.com/IntelligentSoftwareSystems/Galois/pull/200: I can now reproduce these crashes on an Ubuntu 18.04 VM, just not on our development machines. The cause is still not clear.

insertinterestingnamehere:
[...]

PR #199, opened 2020-05-07
Renaming CMake options and preprocessor flags, and fix unused function parameters in lonestar distributed

roshandathathri:


roshandathathri:
@ddn0 I noticed you added GPU tests on CI. Not sure whether my changes includes or breaks your changes.

ddn0:
> I noticed you added GPU tests on CI. Not sure whether my changes includes or breaks your changes.

Thanks for the heads up! I suppose it will depend on which PR lands first, but I don't mind waiting and rebasing on top of your changes.

ddn0:
I assume you are doing #153 with a different PR?

roshandathathri:
Yes, working on it.

roshandathathri:
Fixed the build and renamed the preprocessor defines. Builds fine on UT machine. Tests should pass. I'm testing them on UT machines as well.

ddn0:
> #pragma once instead #ifndef and #define

I have a slight preference for `#ifndef` because it is more conventional, although it is true that gcc and clang and msvc support `#pragma once`.

More generally, if we are going to do this, I'd prefer picking a style and using it consistently for all the code.

Also, with respect to style, I think following established guidelines would be nice. For example:

  https://google.github.io/styleguide/cppguide.html#The__define_Guard

I don't think the Google style guide is particularly amazing or anything, but where things are a matter of taste, I'd prefer things to be settled one way or another so we can devote our energy on more substantial things.

roshandathathri:
Clang format done. Tests pass on UT machines.

roshandathathri:
> I have a slight preference for `#ifndef` because it is more conventional, although it is true that gcc and clang and msvc support `#pragma once`.

I'd prefer pragma once because it is easier to do but I don't mind either. 



> I'd prefer picking a style and using it consistently for all the code.

Agreed. Both styles are widely used in our code. I might have been the main user for "pragma once" - a lot of distributed and gpu code has it. We need to fix this later.

insertinterestingnamehere:
Either is fine. `#pragma once` is widely supported by the compilers we support and could ever support in the foreseeable future (e.g. MSVC). The macros are technically the standard way of doing it, but that's really just a technicality at this point.

roshandathathri:
All failures are unrelated to changes in this PR but they are related to aigRewriting - noted in #132. 

I'm fine with merging this PR once @l-hoang reviews it.

libcusp/include/galois/graphs/MiningPartitioner.h
```diff
@@ -1131,7 +1132,7 @@ class MiningGraph : public DistGraph<NodeTy, EdgeTy> {
         uint64_t cur = *graph.edge_begin(lsrc, galois::MethodFlag::UNPROTECTED);
         uint64_t cur_end = *graph.edge_end(lsrc);
         assert((cur_end - cur) == gdst_vec.size());
-        deserializeEdges(graph, rb, gdst_vec, cur, cur_end);
+        deserializeEdges(graph, gdst_vec, cur, cur_end);
```
l-hoang, line 1135: I'm assuming this change is correct; variable changed completely.

roshandathathri: It's an unused function argument - the change is required to fix the build.

libcusp/include/galois/graphs/MiningPartitioner.h
```diff
@@ -1154,9 +1155,8 @@ class MiningGraph : public DistGraph<NodeTy, EdgeTy> {
   }
 
   template <typename GraphTy>
-  void deserializeEdges(GraphTy& graph, galois::runtime::RecvBuffer& b,
-                        std::vector<uint64_t>& gdst_vec, uint64_t& cur,
-                        uint64_t& cur_end) {
+  void deserializeEdges(GraphTy& graph, std::vector<uint64_t>& gdst_vec,
+                        uint64_t& cur, uint64_t& cur_end) {
```
roshandathathri, line 1159: See this @l-hoang

l-hoang: Acknowledged.

lonestar/libdistbench/include/DistBenchStart.h
```diff
@@ -227,27 +213,28 @@ template <typename NodeData, typename EdgeData, bool iterateOutEdges = true>
...
   Substrate* s;
 
-#ifdef __GALOIS_HET_CUDA__
+#ifdef GALOIS_ENABLE_GPU
   internal::heteroSetup(scaleFactor);
-  g = loadDGraph<NodeData, EdgeData, iterateOutEdges>(scaleFactor, cuda_ctx);
-#else
-  g = loadDGraph<NodeData, EdgeData, iterateOutEdges>(scaleFactor);
 #endif
+  g = loadDGraph<NodeData, EdgeData, iterateOutEdges>(scaleFactor);
```
l-hoang, line 230: Correct me if wrong: cuda_ctx now passed in via distGraphInit instead of loadDGraph. Much cleaner. 👍

roshandathathri: Yes. Loading from file is not related to GPU.

PR #203, opened 2020-05-08
Remove unused profilers

### LINKED ISSUE #186: RFC: Remove HPCTK
If we are only concerned with Linux profiling, `perf` gives us reasonable support for performance counters though not confined to an application-specific region, which is probably preferable for end-to-end performance debugging though. This is probably a good replacement for any PAPI use cases.

The HPCTK support doesn't seem too deep. Does anyone use this?

VTune has reasonable profiling tools but do we need application-specific timing regions, which requires linking to their library?

ddn0:
Closes #186


roshandathathri:
LGTM. I have also requested @yishanlu to review.

@ddn0 can you please check the other CMake flags/options and rename them as appropriate?

ddn0:
Filed a separate issue for renaming the rest of the options #204

roshandathathri:
Thanks, Donald!

ddn0:
The discussion on removing PAPI has stalled, so I've kept it in this PR.

roshandathathri:
I'm voting to keep PAPI but I'm fine if it is dropped.

PR #207, opened 2020-05-08
re-organize lonestar/mining

chenxuhao:
Put mining apps in two separate folders:
lonestar/mining/cpu
lonestar/mining/gpu


insertinterestingnamehere:
Looks great. Thanks!

PR #211, opened 2020-05-11
Allow Locally Disabling -Werror

insertinterestingnamehere:
This adds macros that can be used to turn off `-Werror` for including external headers. Here's a usage example:

```
GALOIS_ALLOW_WARNINGS
#include "external_header_causing_warnings.hpp"
GALOIS_END_ALLOW_WARNINGS
```

### DIFF
--- a/libgalois/include/galois/config.h.in
+++ b/libgalois/include/galois/config.h.in
@@ -19,6 +19,17 @@ _Pragma("GCC diagnostic warning \"-Wextra\"")
 #define GALOIS_END_ALLOW_WARNINGS
 #endif
 
+#if defined(__GNUC__)
+#define GALOIS_IGNORE_WARNINGS               \
+_Pragma("GCC diagnostic push")               \
+_Pragma("GCC diagnostic ignored \"-Wall\"")  \
+_Pragma("GCC diagnostic ignored \"-Wextra\"")
+#define GALOIS_END_IGNORE_WARNINGS _Pragma("GCC diagnostic pop")
+#else
+#define GALOIS_IGNORE_WARNINGS
+#define GALOIS_END_IGNORE_WARNINGS
+#endif
+
 // Macro to suppress compiler warnings that a variable is set but unused.
 // This warning is buggy in gcc 7.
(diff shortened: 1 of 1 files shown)


insertinterestingnamehere:
Note: I could also add macros to ignore all warnings for an include. Just allowing the warnings adds to the build noise while entirely ignoring them is quieter but also seems a bit dangerous. Does anyone have any preferences?

ddn0:
Is there a specific class of warning that is difficult to resolve? Perhaps we should just ignore that as a class (`-Wno-XYZ`) depending on the severity of the warning? Or only make `-Werror` mandatory in CI for certain build configurations? Or just the library code but not the applications? At a certain point, the cure becomes worse than the disease.

If we decide to go forward if the proposed approach, I'd just ignore all warnings since this is a last resort when we cannot control our dependencies. Constantly reminding developers by still emitting the warnings seems counterproductive.

insertinterestingnamehere:
Good question. I can make macros that selectively disable specific warnings if that's what's needed. I'm not the one interacting with the external code though, so I'm not the best one to answer which warnings need to be suppressed. This is really just my generic answer to multiple others who have said "What do we do about external code?" I can change it to ignore instead of just allow warnings though. Really the effect we want is to just not have to worry about upstream code.

roshandathathri:
Changes LGTM.

I agree with Donald that it would be better to ignore specific warning.

ddn0:
> Really the effect we want is to just not have to worry about upstream code.

I'd vote for a blanket ignore and squelch or just allow applications opt into -Werror. There is a limit on how much we can push application writers to change their code.

insertinterestingnamehere:
Currently app owners can easily opt out of `-Werror` by just adding `add_compile_options("$<$<COMPILE_LANGUAGE:CXX>:-Wno-error>")` in their `CMakeLists.txt`. I prefer this over having `-Werror` off by default, but I'm open to other opinions here too. Maintaining things so that `-Werror` can be uniformly applied takes quite a bit of effort, but it also helps us avoid a other potential issues.

PR #214, opened 2020-05-12
TC-Dist-CPU + TC-Dist make test

nicelhc13:
Same as the title.


nicelhc13:
I am reviewing the above errors. 
This was because rmat15.csgr mostly. I tested (make test) with the input locally and it worked.
Unused variable is also not related to this app, but gluon.

insertinterestingnamehere:
I'm rerunning the builds. Does anyone more familiar with the app want to review this?

lonestar/analytics/distributed/tc/tc.cpp
```diff
@@ -61,36 +72,74 @@ template <bool async>
...
+      Graph::edge_iterator p2Begin = graph->edge_begin(p2);
+      Graph::edge_iterator p2End   = graph->edge_end(p2);
+      Graph::edge_iterator p1p     = p1Begin;
+      Graph::edge_iterator p2p     = p2Begin;
+      uint32_t p1Dest, p2Dest;
+
+      while (p1p < p1End && p2p < p2End) {
+        p1Dest           = graph->getEdgeDst(p1p);
+        p2Dest           = graph->getEdgeDst(p2p);
+        int32_t nodeDiff = p1Dest - p2Dest;
```
l-hoang, line 129: I don't particularly like using signed ints to detect termination: problem is that overflow is a concern once the destinations become big enough (e.g. what happens if p1Dest is 0 and p2Dest is 2^31 + 1? you would get an overflow issue). I would change the logic to not have to use arithmetic on the destinations instead if at all possible.

lonestar/analytics/distributed/tc/tc.cpp
```diff
@@ -61,36 +72,74 @@ template <bool async>
...
+
+      Graph::edge_iterator p1Begin = graph->edge_begin(p1);
+      Graph::edge_iterator p1End   = graph->edge_end(p1);
+      Graph::edge_iterator p2Begin = graph->edge_begin(p2);
+      Graph::edge_iterator p2End   = graph->edge_end(p2);
+      Graph::edge_iterator p1p     = p1Begin;
+      Graph::edge_iterator p2p     = p2Begin;
+      uint32_t p1Dest, p2Dest;
+
+      while (p1p < p1End && p2p < p2End) {
```
l-hoang, line 126: See the ordered count algorithm in Lonestar CPU to get an idea of how this loop can be structured as a nested for loop + how to check termination. In my opinion, it's a more understandable way of writing this loop.

lonestar/analytics/distributed/tc/tc.cpp
```diff
@@ -115,28 +164,25 @@ int main(int argc, char** argv) {
...
-  std::tie(hg, syncSubstrate) = distGraphInitialization<NodeData, void>();
-#endif
-
   std::string timer_str("SortEdgesGPU");
   galois::StatTimer edgeSortTime("SortEdgesGPU", regionname);
   edgeSortTime.start();
   sort_cuda(cuda_ctx);
   edgeSortTime.stop();
+#else
+  std::tie(hg, syncSubstrate) = distGraphInitialization<NodeData, void>();
```
l-hoang, line 173: Pass in "false" to distGraphInitialization. It makes it so proxy edge metadata isn't created for sync, which will take up more memory than necessary.

lonestar/analytics/distributed/tc/tc.cpp
```diff
@@ -115,28 +164,25 @@ int main(int argc, char** argv) {
...
```
l-hoang, line 174: Please expose a function to the user in MiningGraph that they can call to sort the graph by destination similar to how sort_cuda is currently called. Note that if the mirror edge synchronization was required this sort would HAVE to be done in construction itself rather than after construction, but right now since the mirror edge sync isn't being used in any app it should be fine.

i.e. See line 332 in MiningPartitioner.h: make base_DistGraph::sortEdges() something the user can call.

lonestar/analytics/distributed/tc/tc.cpp
roshandathathri: Agree. But for now, let's be consistent with what GPU and lonestar cpu is doing.

lonestar/analytics/distributed/tc/tc.cpp
roshandathathri: "sort" is misleading - please have a more descriptive name for both GPU and CPU functions to sort like sortNodesByDegree

lonestar/analytics/distributed/tc/tc.cpp
l-hoang: Lonestar CPU is doing it differently actually: I made a comment below about checking the ordered count algorithm.

lonestar/analytics/distributed/tc/tc.cpp
l-hoang: In this case we're not actually sorting the nodes by degree: it's just sorting edge destinations.

Introducing the sort by degree functionality that ordered count is using into this code would be a nice idea, but in that case it would have to be done before partitioning occurs or done very cleverly on the partitioned graphs on each host such that that LID-GID handshake doesn't get messed up.

l-hoang: ....actually, we don't even need LID-GID handshake. 🤔 
The only sync that occurs is DGAccumulator.

This is something we can look into later.

Re: original point, sortEdgesByDest would be a good name.

roshandathathri: LID-GID mapping is simple to maintain and is not required in this case. 

Each partition can be sort its nodes by degree.

lonestar/analytics/distributed/tc/tc.cpp
```diff
@@ -38,12 +38,23 @@
...
+int personality = CPU;
 #endif
 
 constexpr static const char* const regionname = "TC";
 
 namespace cll = llvm::cl;
 
+enum Exec { Sync, Async };
+
+static cll::opt<Exec> execution(
```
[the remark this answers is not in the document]
nicelhc13: Question: the reason why I put was b/c of make test. Make test passes --exec option. Pagerank also put this dummy option. If we want to remove this, then we should remove make test option, and pagerank's too.

lonestar/analytics/distributed/tc/tc.cpp
nicelhc13: I agree with both of you. Any option works for me. I just tried to make it consistent with GPU codes. So what is your conclusion? If I use unsigned, then GPU codes should be updated, too.

lonestar/analytics/distributed/tc/tc.cpp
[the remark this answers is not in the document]
roshandathathri: You can set NO_ASYNC option for make test. Check bc/CMakeLists.txt

lonestar/analytics/distributed/tc/tc.cpp
l-hoang: GPU has to do things differently due to the threat of warp divergence: CPU and GPU don't have to be consistent.

I would make it consistent with ordered count in Lonestar CPU.

lonestar/analytics/distributed/tc/tc.cpp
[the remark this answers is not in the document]
nicelhc13: Thank you

lonestar/analytics/distributed/tc/tc.cpp
nicelhc13: I see. Thank you.

roshandathathri: Agreed

lonestar/analytics/distributed/tc/tc.cpp
```diff
@@ -61,41 +62,69 @@ template <bool async>
...
-    uint64_t total_triangles = num_triangles.reduce();
+    uint64_t total_triangles = numTriangles.reduce();
     if (galois::runtime::getSystemNetworkInterface().ID == 0) {
       galois::gPrint("Total number of triangles ", total_triangles, "\n");
     }
   }
+
+  void operator()(GNode p1) const {
+    size_t numTriangles_local = 0;
+    for (auto it_p1_1 : graph->edges(p1)) {
```
l-hoang, line 105: minor comment here: the variable names in this loop are quite hard to understand. Being more descriptive with them instead of using p<number> would make it (1) easier to distinguish them among each other (2) easier to understand what they're used for.

nicelhc13: Thanks. Maybe v/vv/w used for shared tc cpu would be better than the current p1/p2/p3. I will update it soon.

libcusp/include/galois/graphs/DistributedGraph.h
```diff
@@ -1006,6 +994,18 @@ class DistGraph {
     galois::gDebug("Deallocating CSR in DistGraph");
     graph.deallocate();
   }
+
+  /**
+   * Sort the underlying LC_CSR_Graph by ID (destinations)
+   */
+  void sortEdges() {
```
roshandathathri, line 1001: More descriptive name please - sortEdgesByDestination() is fine but technically it is sortEdgesOfNodesByDestination

lonestar/analytics/distributed/tc/tc.cpp
```diff
@@ -115,28 +144,28 @@ int main(int argc, char** argv) {
...
   std::tie(hg, syncSubstrate) =
       distGraphInitialization<NodeData, void>(&cuda_ctx, false);
-#else
-  std::tie(hg, syncSubstrate) = distGraphInitialization<NodeData, void>();
-#endif
-
   std::string timer_str("SortEdgesGPU");
   galois::StatTimer edgeSortTime("SortEdgesGPU", regionname);
   edgeSortTime.start();
   sort_cuda(cuda_ctx);
```
roshandathathri, line 150: Same comment. More descriptive name please - use same name sortEdgesByDestination_cuda()

roshandathathri:
Except the name for the sorting function, LGTM

insertinterestingnamehere:
Merging. We'll fix the unused variable separately.

PR #215, opened 2020-05-12
Parallel relabelling for triangle counting

l-hoang:
Changed the sorting routine used by triangle counting to be more efficient. Before, it was basically single threaded: more parallelism has been introduced into the procedure.

One thing that may help further is if std::sort parallelism is enabled (if it isn't). GAP is very likely doing so as it enables OpenMP.

https://stackoverflow.com/questions/28520720/c-parallel-sort

This also includes a change to LC_CSR_Graph to expose the local range initialization to allow threads to iterate over a more balanced workload (and required if you want to directly call galois::iterate on an LC_CSR_Graph object).

### DIFF
--- a/libgalois/include/galois/ParallelSTL.h
+++ b/libgalois/include/galois/ParallelSTL.h
@@ -305,6 +305,60 @@ std::enable_if_t<!std::is_scalar<internal::Val_ty<I>>::value> destroy(I first,
 template <class I>
 std::enable_if_t<std::is_scalar<internal::Val_ty<I>>::value> destroy(I, I) {}
 
+/**
+ * Does a partial sum from first -> last and writes the results to the d_first
+ * iterator.
+ */
+template <class InputIt, class OutputIt>
+OutputIt partial_sum(InputIt first, InputIt last, OutputIt d_first) {
+  using ValueType = typename std::iterator_traits<InputIt>::value_type;
+
+  size_t sizeOfVector = std::distance(first, last);
+
+  const size_t blockSize = 1 << 20;
+  // optimization possible here: if num blocks = 1, then no need for 2 passes
+  const size_t numBlocks = (sizeOfVector + blockSize - 1) / blockSize;
+
+  std::vector<ValueType> localSums(numBlocks);
+
+  // get the block sums
+  galois::do_all(
+      galois::iterate((size_t)0, numBlocks), [&](const size_t& block) {
+        ValueType lsum = 0;
+        // loop over block
+        size_t blockEnd = std::min((block + 1) * blockSize, sizeOfVector);
+        for (size_t i = block * blockSize; i < blockEnd; i++) {
+          lsum += *(first + i);
+        }
+        localSums[block] = lsum;
+      });
+
+  // bulkPrefix[i] holds the starting sum of a particular block i
+  std::vector<ValueType> bulkPrefix(numBlocks);
+  // first block starts at 0; rest start at sum of all previous blocks
+  ValueType total = 0;
+  for (size_t block = 0; block < numBlocks; block++) {
+    bulkPrefix[block] = total;
+    total += localSums[block];
+  }
+
+  galois::do_all(
+      galois::iterate((size_t)0, numBlocks), [&](const size_t& block) {
+        // start with this block's prefix up to this point
+        ValueType localTotal = bulkPrefix[block];
+        size_t blockEnd      = std::min((block + 1) * blockSize, sizeOfVector);
+
+        for (size_t i = block * blockSize; i < blockEnd; i++) {
+          // add current element first, then save
+          localTotal += *(first + i);
+          *(d_first + i) = localTotal;
+        }
+      });
+
+  // return the iterator past the last element written
+  return d_first + sizeOfVector;
+}
+
 } // end namespace ParallelSTL
 } // end namespace galois
 #endif
--- a/libgalois/include/galois/ParallelSTL.h
+++ b/libgalois/include/galois/ParallelSTL.h
@@ -27,6 +27,7 @@
 #include "galois/Reduction.h"
 #include "galois/Traits.h"
 #include "galois/UserContext.h"
+#include "galois/Threads.h"
 #include "galois/worklists/Chunk.h"
 
 namespace galois {
@@ -315,9 +316,10 @@ OutputIt partial_sum(InputIt first, InputIt last, OutputIt d_first) {
 
   size_t sizeOfVector = std::distance(first, last);
 
-  const size_t blockSize = 1 << 20;
   // optimization possible here: if num blocks = 1, then no need for 2 passes
-  const size_t numBlocks = (sizeOfVector + blockSize - 1) / blockSize;
+  const size_t numBlocks = galois::getActiveThreads();
+  const size_t blockSize = (sizeOfVector + numBlocks - 1) / numBlocks;
+  assert(numBlocks * blockSize >= sizeOfVector);
 
   std::vector<ValueType> localSums(numBlocks);
 
--- a/libgalois/include/galois/ParallelSTL.h
+++ b/libgalois/include/galois/ParallelSTL.h
@@ -316,49 +316,65 @@ OutputIt partial_sum(InputIt first, InputIt last, OutputIt d_first) {
 
   size_t sizeOfVector = std::distance(first, last);
 
-  // optimization possible here: if num blocks = 1, then no need for 2 passes
-  const size_t numBlocks = galois::getActiveThreads();
-  const size_t blockSize = (sizeOfVector + numBlocks - 1) / numBlocks;
-  assert(numBlocks * blockSize >= sizeOfVector);
+  // only bother with parallel execution if vector is larger than some size
+  if (sizeOfVector >= 1024) {
+    const size_t numBlocks = galois::getActiveThreads();
+    const size_t blockSize = (sizeOfVector + numBlocks - 1) / numBlocks;
+    assert(numBlocks * blockSize >= sizeOfVector);
 
-  std::vector<ValueType> localSums(numBlocks);
+    std::vector<ValueType> localSums(numBlocks);
 
-  // get the block sums
-  galois::do_all(
-      galois::iterate((size_t)0, numBlocks), [&](const size_t& block) {
-        ValueType lsum = 0;
-        // loop over block
-        size_t blockEnd = std::min((block + 1) * blockSize, sizeOfVector);
-        for (size_t i = block * blockSize; i < blockEnd; i++) {
-          lsum += *(first + i);
-        }
-        localSums[block] = lsum;
-      });
+    // get the block sums
+    galois::do_all(
+        galois::iterate((size_t)0, numBlocks), [&](const size_t& block) {
+          // block start can extend past sizeOfVector if doesn't divide evenly
+          size_t blockStart = std::min(block * blockSize, sizeOfVector);
+          size_t blockEnd   = std::min((block + 1) * blockSize, sizeOfVector);
+          assert(blockStart <= blockEnd);
 
-  // bulkPrefix[i] holds the starting sum of a particular block i
-  std::vector<ValueType> bulkPrefix(numBlocks);
-  // first block starts at 0; rest start at sum of all previous blocks
-  ValueType total = 0;
-  for (size_t block = 0; block < numBlocks; block++) {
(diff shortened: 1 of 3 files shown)


lonestar/analytics/cpu/triangles/Triangles.cpp
```diff
@@ -354,46 +354,98 @@ void edgeIteratingAlgo(Graph& graph) {
...
+      // save degree, which is pair.first
+      newPrefixSum[index] = dnPairs[index].first;
+      // save mapping; original index is in .second, map it to current index
+      oldToNewMapping[dnPairs[index].second] = index;
+    },
+    galois::loopname("CreateRemappingGetPrefixSum")
+  );
+
+  // get prefix sum
+  for (size_t i = 1; i < numGraphNodes; i++) {
```
roshandathathri, line 394: This should be parallel. We might have a parallel prefix sum routine in our library. Check with @chenxuhao. Otherwise, use on_each and parallelize it (you can merge it with the previous do_all as well).

lonestar/analytics/cpu/triangles/Triangles.cpp
```diff
@@ -354,46 +354,98 @@ void edgeIteratingAlgo(Graph& graph) {
...
+    [&] (size_t nodeID) {
+      size_t nodeDegree = std::distance(initial.edge_begin(nodeID),
+                                        initial.edge_end(nodeID));
+      dnPairs[nodeID] = DegreeNodePair(nodeDegree, nodeID);
+    },
+    galois::loopname("CreateDegreeNodeVector")
+  );
+
+  // sort by degree (first item)
+  std::sort(dnPairs.begin(), dnPairs.end(), std::greater<DegreeNodePair>());
```
roshandathathri, line 378: Parallelize this using galois::ParallelSTL::sort()

lonestar/analytics/cpu/triangles/Triangles.cpp
```diff
@@ -354,46 +354,98 @@ void edgeIteratingAlgo(Graph& graph) {
...
-
-  /* Sort by degree:
-   *  DegreeLess: Sorts in the ascending order of node degrees
-   *  DegreeGreater: Sorts in the descending order of the node degrees
-   *
-   *  The order of sorting has a huge impact on performance
-   *  For this algorithm, sorting in descending order delivers the
-   *  best performance due to the way ties are broken.
-   */
-  galois::ParallelSTL::sort(nodes.begin(), nodes.end(),
```
roshandathathri, line 377: New sort has the same signature as this, right?

lonestar/analytics/cpu/triangles/Triangles.cpp
```diff
@@ -403,12 +455,12 @@ void readGraph(Graph& graph) {
     std::string triangleFilename = inputFilename + ".triangles";
     std::ifstream triangleFile(triangleFilename.c_str());
     if (!triangleFile.good()) {
-      // triangles doesn't already exist, create it
+      // triangles doesn't already exist
       galois::StatTimer Trelabel("GraphRelabelTimer");
-      galois::gPrint("WARNING: Sorted graph does not exist; Relabelling and "
-                     "Creating a sorted graph\n");
+      galois::gPrint("WARNING: Sorted graph (.triangles) does not exist; Relabelling and "
```
roshandathathri, line 460: I'm in favor of dropping the option to load a sorted graph. It keeps this simple and consistent among benchmarks.

lonestar/analytics/cpu/triangles/Triangles.cpp
l-hoang: ack

lonestar/analytics/cpu/triangles/Triangles.cpp
l-hoang: yes

lonestar/analytics/cpu/triangles/Triangles.cpp
l-hoang: No objections. @gurbinder533?

lonestar/analytics/cpu/triangles/Triangles.cpp
l-hoang: pinged Xuhao elsewhere, will see what he says.

lonestar/analytics/cpu/triangles/Triangles.cpp
l-hoang: I do have one comment though.

I mentioned this to you before Roshan, but we really should implement parallel input processing independent of however many threads the user decides to run the actual application for itself. I can see this being an issue if someone wants to check scaling and they're stuck with a single-threaded reordering scheme.

Maybe not for this upcoming release, but it should happen at some point.

roshandathathri: Good point. 

Not required for this release. But in the future, we can add a hidden option to specify number of threads for file loading.

lonestar/analytics/cpu/triangles/Triangles.cpp
l-hoang: resolved in local commit

lonestar/analytics/cpu/triangles/Triangles.cpp
l-hoang: Will be moving the parallel prefix sum to ParallelSTL as a function.

lonestar/analytics/cpu/triangles/Triangles.cpp
l-hoang: Dropped in local commit. If Gurbinder objects he may make his objections known later. :P

l-hoang:
Added partial_sum to parallel stl + am using it in relabelling. @roshandathathri please check.

libgalois/include/galois/ParallelSTL.h
```diff
@@ -305,6 +305,60 @@ std::enable_if_t<!std::is_scalar<internal::Val_ty<I>>::value> destroy(I first,
...
+ */
+template <class InputIt, class OutputIt>
+OutputIt partial_sum(InputIt first, InputIt last, OutputIt d_first) {
+  using ValueType = typename std::iterator_traits<InputIt>::value_type;
+
+  size_t sizeOfVector = std::distance(first, last);
+
+  const size_t blockSize = 1 << 20;
+  // optimization possible here: if num blocks = 1, then no need for 2 passes
+  const size_t numBlocks = (sizeOfVector + blockSize - 1) / blockSize;
```
roshandathathri, line 320: I think the number of blocks should be the same as the number of threads. There's no need to load balance here and the iterations of the serial loop would be reduced.

l-hoang: Doable.

My concern is that for trivially small vectors there really isn't a point going multi-threaded since there is overhead with parallelism involved, but I think in most cases it probably big enough of an issue to design around that.

One more comment I forgot to make: the std::partial_sum interface makes it so you write to a new vector (the output vector), so doing this actually adds another O(n) memory overhead. It probably isn't too much of a deal considering relabeling requires having 2 graphs in memory anyways.

l-hoang: Made the change; do check to make sure logic is fine.

l-hoang:
Should be ready to merge once Roshan gives OK.

libgalois/include/galois/ParallelSTL.h
```diff
@@ -305,6 +306,61 @@ std::enable_if_t<!std::is_scalar<internal::Val_ty<I>>::value> destroy(I first,
...
+ * Does a partial sum from first -> last and writes the results to the d_first
+ * iterator.
+ */
+template <class InputIt, class OutputIt>
+OutputIt partial_sum(InputIt first, InputIt last, OutputIt d_first) {
+  using ValueType = typename std::iterator_traits<InputIt>::value_type;
+
+  size_t sizeOfVector = std::distance(first, last);
+
+  // optimization possible here: if num blocks = 1, then no need for 2 passes
```
roshandathathri, line 319: This is simple to enforce. Just check if sizeOfVector <= 1024 (use the same value as sort in the file), then call std::partial_sum() and return.

libgalois/include/galois/ParallelSTL.h
```diff
@@ -305,6 +306,61 @@ std::enable_if_t<!std::is_scalar<internal::Val_ty<I>>::value> destroy(I first,
...
+  std::vector<ValueType> localSums(numBlocks);
+
+  // get the block sums
+  galois::do_all(
+      galois::iterate((size_t)0, numBlocks), [&](const size_t& block) {
+        ValueType lsum = 0;
+        // loop over block
+        size_t blockEnd = std::min((block + 1) * blockSize, sizeOfVector);
+        for (size_t i = block * blockSize; i < blockEnd; i++) {
+          lsum += *(first + i);
```
roshandathathri, line 333: It's better to partially accumulate the blocks here. In fact, I would prefer if you can just std::partial_sort() on each block in each thread.

libgalois/include/galois/ParallelSTL.h
```diff
@@ -305,6 +306,61 @@ std::enable_if_t<!std::is_scalar<internal::Val_ty<I>>::value> destroy(I first,
...
+
+  galois::do_all(
+      galois::iterate((size_t)0, numBlocks), [&](const size_t& block) {
+        // start with this block's prefix up to this point
+        ValueType localTotal = bulkPrefix[block];
+        size_t blockEnd      = std::min((block + 1) * blockSize, sizeOfVector);
+
+        for (size_t i = block * blockSize; i < blockEnd; i++) {
+          // add current element first, then save
+          localTotal += *(first + i);
```
roshandathathri, line 355: Just add localTotal to all elements of the block in each thread. You can use some std function to that - I think std transform is suitable here.

libgalois/include/galois/ParallelSTL.h
l-hoang: ack

libgalois/include/galois/ParallelSTL.h
l-hoang: std::transform requires another output vector; think it's easier to just use a for loop and add localTotal to all elements.

libgalois/include/galois/ParallelSTL.h
l-hoang: ack

libgalois/include/galois/ParallelSTL.h
roshandathathri: It's fine either way but are you sure the output vector cannot be the same as the input vector?

l-hoang: Just checked: it can be.
I honestly think just writing a for loop is easier to understand though; std::transform would involve defining a lambda.

roshandathathri: I would prefer to use std functions like copy or transform because they are likely to use some optimizations like vector instructions.

l-hoang:
@roshandathathri Revised partial_sum logic: please check.

libgalois/include/galois/ParallelSTL.h
l-hoang: pushing in a forced-push momentarily

libgalois/include/galois/ParallelSTL.h
```diff
@@ -305,6 +306,74 @@ std::enable_if_t<!std::is_scalar<internal::Val_ty<I>>::value> destroy(I first,
...
+ */
+template <class InputIt, class OutputIt>
+OutputIt partial_sum(InputIt first, InputIt last, OutputIt d_first) {
+  using ValueType = typename std::iterator_traits<InputIt>::value_type;
+
+  size_t sizeOfVector = std::distance(first, last);
+
+  // only bother with parallel execution if vector is larger than some size
+  if (sizeOfVector >= 1024) {
+    // optimization possible here: if num blocks = 1, then no need for 2 passes
```
roshandathathri, line 321: Remove comment and make it part of the if statement

l-hoang:
removed comment + using std::transform in latest commit

l-hoang:
Arch failure is a SProute failure it looks like: ready to merge whenever.

insertinterestingnamehere:
Yes, the Arch failure is hopefully already fixed in master. Will merge shortly.

insertinterestingnamehere:
Yah, the Travis CI failures are unrelated and already fixed in master as well. Merging. Thanks!

Issue #219, opened 2020-05-13
Personality setting in distributed apps


l-hoang:
Roshan was going through the code and noticed this (and it's there in the TC code I'm currently reviewing):

#ifdef GALOIS_ENABLE_GPU
#include "bfs_push_cuda.h"
struct CUDA_Context* cuda_ctx;
#else
enum { CPU, GPU_CUDA };
int personality = CPU;
#endif

This as well:
  if (personality == GPU_CUDA) {
#ifdef GALOIS_ENABLE_GPU
      std::string impl_str(
          syncSubstrate->get_run_identifier("InitializeGraph_"));
      galois::StatTimer StatTimer_cuda(impl_str.c_str(), regionname);
      StatTimer_cuda.start();
      InitializeGraph_allNodes_cuda(infinity, src_node, cuda_ctx);
      StatTimer_cuda.stop();
#else
      abort();
#endif

Apparently this change was done by @ddn0. The problem with this change is that personality doesn't really mean anything if ENABLE_GPU isn't there. The purpose of the personality flag is to enable heterogeneous execution, i.e., some hosts run with CPUs, others with GPUs. It doesn't mean much if the GPUs aren't enabled because then your only setting is CPUs anyways.

Is there a reason the dist apps were changed to have this new pattern?

ddn0:
> It doesn't mean much if the GPUs aren't enabled because then your only setting is CPUs anyways.

That was certainly the intent. To reduce the number of codepaths and if there is going to be some configure-time dispatch, it should follow the syntax of C++ rather than of the C preprocessor.

If the naming is non-intuitive, it might make sense to rename things. But what we should avoid is sprinkling a bunch of ifdefs everywhere, includes that happen in the middle of a file, and any use of ifdefs that can be accomplished with C++ code instead.

roshandathathri:
@ddn0 I did not quite understand. Adding personality when GPU is disabled is not reducing the number of #ifdefs but adding more lines of code. Is this the solution you're proposing or is this a step towards a better solution?

ddn0:
> Adding personality when GPU is disabled is not reducing the number of #ifdefs but adding more lines of code.

My reasoning was using the CPU when you don't GPU support should probably look the same as using the CPU when you do have GPU support.

In general, I'd aim for increasing reusability and simplicity which can be different than purely reducing the number of lines of code.

ddn0:
As an example of the kind of code structure that I think we should avoid:

[...]

Issue #195, opened 2020-05-07
Internal compiler error in CUDA build


ddn0:
Observed with CUDA 10.1 and 10.2

```
/data/lonestar/analytics/gpu/pta/andersen.cu(1379): internal error: assertion failed: conv_glvalue_expr_to_prvalue: bad expr (/dvs/p4/build/sw/rel/gpu_drv/r418/trd418_98/drivers/compiler/edg/edg_5.0/src/exprutil.c, line 20215)


1 catastrophic error detected in the compilation of "/tmp/tmpxft_00001c6f_00000000-4_pta.cpp4.ii".
````

Example: https://app.circleci.com/pipelines/github/IntelligentSoftwareSystems/Galois/384/workflows/ea8967fe-9e9f-4bea-9bb3-7eb4be31418d/jobs/7626/steps

vishweshjatala:
This test builds correctly on our local machine. It seems to be failing only on Ubuntu18.04. I don't have access to Ubuntu 18.04. Is there a way we can try to  reproduce it?

ddn0:
See https://github.com/IntelligentSoftwareSystems/Galois/issues/194

vishweshjatala:
CUDA builds are passing now. I created a separate issue (https://github.com/IntelligentSoftwareSystems/Galois/issues/228) for Fedora-clang build error. We can close this issue if it is fine.

vishweshjatala:
PR https://github.com/IntelligentSoftwareSystems/Galois/issues/194 is merged. Closing it.

PR #220, opened 2020-05-14
Structured output

ddn0:
Add a command line option to write results `--outputLocation=string`/`--output`.

Downstream, we will be writing these out as parquet files probably, but there is no reason everyone can't benefit from the common extraction code. The output code is outside any timing region so this shouldn't affect benchmark results.

We intend to support remote storage backends like S3, which is why commands do not currently create the output directory if it is missing.

The output API (`writeOutput`) implicitly assumes that node data can just be concatenated together as there is no parameter for indices. We can technically support sparse node data, but as far as I could tell, the current interfaces do not yet expose this functionality (masterNode iterators are just `counting_iterator`s). So, I'm deferring this work until the graph interfaces/implications can change too.

I briefly considered adding parquet support here but I didn't want to upset the applecart during the release window. Also, the implementation of node data (properties) is still TBD; in particular, we might need to change the graph implementations from the implicit array of structs representation to a struct of arrays to get zero-copy serialization.

I know that there has been some discussion on the appropriate way to dispatch CPU versus GPU code #219. I haven't seen a resolution here, so I just continued the current path.

Tested locally. I had to manually work around #195:

```
ERROR: /home/ddn/w/katana/libgalois/src/FileGraph.cpp:304: No such file or directory: failed opening '/home/ddn/w/katana-build/inputs/Mining/citeseer.csgr'
...
The following tests FAILED:
        408 - run-small1-kcl-6 (Child aborted)
        409 - run-small1-kcl-3 (Child aborted)
        410 - run-small1-kcl-1 (Child aborted)
```

### DIFF
--- a/lonestar/libdistbench/include/DistBench/Output.h
+++ b/lonestar/libdistbench/include/DistBench/Output.h
@@ -1,0 +1,23 @@
+#ifndef GALOIS_DISTBENCH_OUTPUT_H
+#define GALOIS_DISTBENCH_OUTPUT_H
+
+#include <filesystem>
+#include <string>
+#include <vector>
+#include <fstream>
+
+std::string makeOutputFilename(const std::string& outputDir);
+
+template <typename T>
+void writeOutput(const std::string& outputDir, const std::string& /*fieldName*/,
+                 T* values, size_t length) {
+  std::string filename = makeOutputFilename(outputDir);
+
+  std::ofstream outputFile(filename.c_str());
+
+  for (size_t i = 0; i < length; i++) {
+    outputFile << *(values++) << "\n";
+  }
+}
+
+#endif
--- a/lonestar/libdistbench/include/DistBench/Output.h
+++ b/lonestar/libdistbench/include/DistBench/Output.h
@@ -1,9 +1,7 @@
 #ifndef GALOIS_DISTBENCH_OUTPUT_H
 #define GALOIS_DISTBENCH_OUTPUT_H
 
-#include <filesystem>
 #include <string>
-#include <vector>
 #include <fstream>
 
 std::string makeOutputFilename(const std::string& outputDir);
--- a/lonestar/analytics/distributed/README.md
+++ b/lonestar/analytics/distributed/README.md
@@ -99,11 +99,11 @@ Number of threads to use on a single machine excluding a communication thread
 that is used by all of the provided distributed benchmarks. Note that 
 GPUs only use 1 thread (excluding the communication thread).
 
-`-verify`
+`-output` / `-outputLocation=<directory>`
 
-Outputs a file with the result of running the application. For example, 
-specifying this flag on a bfs application will output the shortest distances
-to each node.
+Outputs the result of running the application to a file. For example,
+specifying this flag on a bfs application will output the shortest distances to
+each node.
 
 Running Provided Apps (Distributed Heterogeneous Apps)
 ================================================================================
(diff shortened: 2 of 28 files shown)


ddn0:
It occurs to me that going forward we might just want to replace `verify` with `outputDirectory` because none of the analytics applications have true verification routines anyway. Having a standard output format is a prerequisite for this

lonestar/libdistbench/include/DistBench/Output.h
```diff
@@ -0,0 +1,21 @@
...
+#define GALOIS_DISTBENCH_OUTPUT_H
+
+#include <string>
+#include <fstream>
+
+std::string makeOutputFilename(const std::string& outputDir);
+
+template <typename T>
+void writeOutput(const std::string& outputDir, const std::string& /*fieldName*/,
+                 T* values, size_t length) {
```
roshandathathri, line 11: Why not take std::vector for values?

roshandathathri:
Thanks, Donald. This is a great idea!

> It occurs to me that going forward we might just want to replace verify with outputDirectory because none of the analytics applications have true verification routines anyway. Having a standard output format is a prerequisite for this

I'm in favor of this. "-verify" means different things for shared-memory and distributed-memory apps currently, which is bad. Also, two code paths for printing is also confusing/bad. We should use "-verify" for in-memory verification and "-output" for printing the output. If "-output" is used and "-outputDirectory" is not specified, then we print it to the current directory.

lonestar/libdistbench/include/DistBench/Output.h
ddn0: That was in my original API :D Going to pointers may be a little more future-looking than necessary at the moment, but the idea is that this data may come from a pure buffer (GPU memory, mmap'ed file, etc). And if we go to purely zero copy, it will be difficult for users to encapsulate their data as `std::vector`s.

roshandathathri: Makes sense. Agreed. 

We can avoid copy for GPU but I'm not sure if I have exposed an API for that. I'll check and let you know.

ddn0:
> We should use "-verify" for in-memory verification and "-output" for printing the output.

Agreed. If there isn't too much controversy in this PR, I can do this change here. Otherwise, I can file an issue for this follow on work.

> If "-output" is used and "-outputDirectory" is not specified, then we print it to the current directory.

Your preference is for the two argument version (`--output` and `--outputDirectory`) over the current single argument version (`--outputDirectory`)? Just checking as I have no strong preference myself. If we do go to the two argument version, perhaps `--outputLocation` instead of `--outputDirectory`? My immediate use-case for this is to write-back results to S3 so "location" seems a little more appropriate.

roshandathathri:
> Agreed. If there isn't too much controversy in this PR, I can do this change here. Otherwise, I can file an issue for this follow on work.

I'm fine if you use the same PR for that. @l-hoang and @gurbinder533, confirm whether you're fine moving from "-verify" to "-output" for dist apps.

> Your preference is for the two argument version (`--output` and `--outputDirectory`) over the current single argument version (`--outputDirectory`)?

Yes, but I'm fine with going with one option as well and except the user to specify current directory/location.

> `--outputLocation` instead of `--outputDirectory`?

Yes, that's better.

l-hoang:
Fine with moving to -output: verify doesn't really make sense given that shared-mem uses verify for a completely different purpose, and it's not actually doing any verification (just output printing).

PR #226, opened 2020-05-15
Re-enable aigRewriting tests in CI.

insertinterestingnamehere:


insertinterestingnamehere:
Failing sanitizer build is because of a misaligned memory access at https://github.com/IntelligentSoftwareSystems/Galois/blob/master/lonestar/eda/cpu/aigRewriting/algorithms/CutPool.cpp#L72. The fix isn't obvious to me yet.

insertinterestingnamehere:
@yishanlu if you get time at some point I'd appreciate confirmation that the fix here looks sane to you. I'll probably merge this very soon if the tests pass though to get it included in the release, but feel free to leave feedback even after this is merged. If the fix isn't quite right, we can backport additional changes.

lonestar/eda/cpu/aig-rewriting/algorithms/CutPool.cpp
```diff
@@ -57,8 +57,8 @@ CutPool::~CutPool() {
 
 inline void CutPool::alloc() {
 
-  this->entriesFree =
-      (char*)malloc((long int)(this->entrySize * this->blockSize));
+  this->entriesFree = (char*)malloc(
+      (long int)(this->entrySize * this->blockSize * sizeof(char*)));
```
yishanlu, line 61: The pointer is of type `char*`, so the allocated space should be using `sizeof(char)`?

lonestar/eda/cpu/aig-rewriting/algorithms/CutPool.cpp
```diff
@@ -69,8 +69,8 @@ inline void CutPool::alloc() {
   char* pTemp = this->entriesFree;
 
   for (int i = 1; i < this->blockSize; i++) {
-    *((char**)pTemp) = pTemp + this->entrySize;
-    pTemp += this->entrySize;
+    *((char**)pTemp) = pTemp + this->entrySize * sizeof(char*);
+    pTemp += this->entrySize * sizeof(char*);
```
yishanlu, line 73: I don't quite get the purpose of this code snippet. Looks like it's trying to initialize some kind of row pointers. Isn't the underlying element size already considered by pointer arithmetic?

lonestar/eda/cpu/aig-rewriting/algorithms/CutPool.cpp
yishanlu: Also `sizeof(char)` is always 1

insertinterestingnamehere: Here's how I understand it, FWIW. Even though the result of `malloc` is getting cast to `char*` here, it's actually later used as a `char**`. That intent wasn't reflected in the `malloc` size or in the computed offset from the base pointer for the buffer. Prior to this there were misaligned writes for type `char*` happening. I think it likely was also overriding portions of previously stored pointers due to these misaligned accesses. I'm surprised the previous version even worked. Unfortunately I know little about the purpose of this bit of code so I'm not entirely certain my fix is right either.

lonestar/eda/cpu/aig-rewriting/algorithms/CutPool.cpp
insertinterestingnamehere: Yah, reiterating what I said above, the main confusion here is that it looks like they want to allocate a buffer of `char*` pointers, but instead of storing that value as a `char*` whoever wrote this opted stored the result from malloc in a `char*` and then later cast it to a `char**` when the needed it. That's where the missing `sizeof(char*)` factor comes from. This code is pretty weird though, so I appreciate you looking at it too.

yishanlu: In my understanding, `CutPool` is managing memory on for cut enumeration in AIG rewriting. This is rewritten mainly in C-style, so may get confused. Note that it is shifting by `entrySize` for every row, so the results are correct, just that it may have non-aligned accesses on some architectures. Your fix essentially bloats up the `entrySize` by `sizeof(char*)`; no semantic changes, but uses more memory for aligned accesses.

insertinterestingnamehere: Okay, I think I see what you're saying. So it *was* actually allocating the right amount, but it wasn't obeying alignment constraints because the right value for `entrySize` isn't necessarily a multiple of 8?

lonestar/eda/cpu/aig-rewriting/algorithms/CutPool.cpp
insertinterestingnamehere: I don't see space for the pointer included in https://github.com/IntelligentSoftwareSystems/Galois/pull/226/files/a6b504f7033ef24e99a086f7c98c8f80d2940084#diff-3c61213a4fd449ea9fd96523fd5b2eb2R41-R46, but then I'm not 100% sure what each of those terms mean.

insertinterestingnamehere: Whoops. Last comment was meant to be on the other thread. I'll copy it to there.

lonestar/eda/cpu/aig-rewriting/algorithms/CutPool.cpp
insertinterestingnamehere: I don't see space for the pointer included in https://github.com/IntelligentSoftwareSystems/Galois/pull/226/files/a6b504f7033ef24e99a086f7c98c8f80d2940084#diff-3c61213a4fd449ea9fd96523fd5b2eb2R41-R46, but then I'm not 100% sure what each of those terms mean.

yishanlu: > Okay, I think I see what you're saying. So it _was_ actually allocating the right amount, but it wasn't obeying alignment constraints because the right value for `entrySize` isn't necessarily a multiple of 8?

Yes, exactly. `entrySize` is defined in `CutPool::CutPool()`, so it is actually allocating space for cuts.

Note that it is allocating `blockSize*entrySize` bytes, in which `entrySize` is the size of a cut and `blockSize` is the number of cuts allocated in one chunk. The loop here tries to embed a linked list of cuts in a contiguously allocated memory.

A better fix would be to round `entrySize` to the multiple of sizeof(char*). This is in `CutPool::CutPool()`.

yishanlu: And I see why the code is written in this way: the memory space is cast to `Cut*` in `CutPool::getMemory()`and back to `char*` in `CurPool::giveBackMemory()`. This would be similar to PerThreadStorage in Galois.

insertinterestingnamehere: Got it. So the pointer size is included in `sizeof(Cut)`, so the right fix is to just actually follow the alignment constraints.

insertinterestingnamehere:
Okay, I've pushed a new fix that works locally. We'll see what the tests think.

insertinterestingnamehere:
Failure is unrelated. Merging.

PR #240, opened 2020-05-21
Clean up LC_CSR_Graph

ddn0:
Remove degree and hyper graph fields, which are specific to graph
specializations.

Delete dead code.


libgalois/include/galois/graphs/LC_CSR_Graph.h
```diff
@@ -677,7 +650,7 @@ class LC_CSR_Graph :
           edgeDst_old[e] = dst;
           // counting outgoing edges in the tranpose graph by
           // counting incoming edges in the original graph
-          __sync_add_and_fetch(&(edgeIndData_temp[dst]), 1);
+          __sync_add_and_fetch(&edgeIndData_temp[dst], 1);
```
insertinterestingnamehere, line 653: Not for this PR, but we should migrate to exclusively using `std::atomic` wherever possible. In many cases we're accidentally getting sequential memory consistency by using the `__sync` prefixed atomic builtins.

PR #242, opened 2020-05-21
Standardize uses of GALOIS_DIE

ddn0:
Error messages should be user actionable if possible. This means that
they should expressed in terms of user-facing abstractions (e.g., files,
command line arguments) and phrased in such a way as to indicate what
the remedy should be.

Along the way, give errors with the same causes the same message, adopt
an initial lower-case for messages, and remove excess developer-facing
context. The line number printed by GALOIS_DIE should be sufficient to
identify the particular GALOIS_DIE statement.

### DIFF
--- a/lonestar/analytics/cpu/kcore/kcore.cpp
+++ b/lonestar/analytics/cpu/kcore/kcore.cpp
@@ -235,8 +235,9 @@ int main(int argc, char** argv) {
   LonestarStart(argc, argv, name, desc, url);
 
   if (!symmetricGraph) {
-    GALOIS_DIE("User did not pass in symmetric graph flag signifying they are "
-               "aware this program needs to be passed a symmetric graph.");
+    GALOIS_DIE("kcore requires a symmetric graph input;"
+               " please use the -symmetricGraph flag "
+               " to indicate the input is a symmetric graph");
   }
 
   //! Some initial stat reporting.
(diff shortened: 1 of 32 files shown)


lonestar/analytics/cpu/kcore/kcore.cpp
```diff
@@ -235,8 +235,7 @@ int main(int argc, char** argv) {
   LonestarStart(argc, argv, name, desc, url);
 
   if (!symmetricGraph) {
-    GALOIS_DIE("User did not pass in symmetric graph flag signifying they are "
-               "aware this program needs to be passed a symmetric graph.");
+    GALOIS_DIE("kcore requires a symmetric graph input");
```
l-hoang, line 238: Prefer the older message in this case as the user does not have to require code/debugging to fix this error: they only have to pass in the -symmetricGraph flag as acknowledgment that they know the graph passed in should be symmetric.

ddn0: Sure!

How about "kcore requires a symmetric graph input; please use the -symmetricGraph flag to indicate the input is a symmetric graph"?

The old message had two issues:
- Referred to the user in the 3rd person, refers to the mental state of the user rather than facts about the input itself and command line options
- Didn't achieve the goal of not looking at code to resolve the issue because it didn't even give the actual name of the flag required

l-hoang: Good point about the previous message's issues. The new one suggested is good.

PR #247, opened 2020-05-22
Re-enable gmetis test in CI.

insertinterestingnamehere:


insertinterestingnamehere:
Weird. Tons of unrelated tests are failing now. Not sure what's going on here yet.

insertinterestingnamehere:
The failures appear to have been caused by a bad inputs download. I was able to download and untar the file manually though, so it's not clear what happened. I'm rerunning the tests to see if the bad download was just a fluke.

Issue #161, opened 2020-04-30
SymmetricGraph option


nicelhc13:
Lonestar has applications which must use symmetric graphs.
I think most of the apps already have the 'symmetricGraph' option which enforce users to specify it, but if there is any app which still does not have it, then we should add it for clarification.

For example, k-truss did not have an option and I added it.

Similarly, we can add -transposeGraph option to the pagerank residual if you guys agree.

ddn0:
+1 for having a uniform command line interface.

A common user question is "what inputs does application X take?" Anything that makes it more clear to the user that one application must take a symmetric graph and another **might** take a directed one would be great!

nicelhc13:
That is good idea. I want to raise this issue on the Thursday discussion. Hi, @insertinterestingnamehere , is Tuesday hard deadline for the release?

insertinterestingnamehere:
Sorry I forgot to bring this one up this morning. We're trying to get all code cleanup and bug fixes done this week so we can just worry about documentation after that. The hard deadline is the end of May, but we have to leave time to work on documentation and also just to give everyone time to test that everything is still working how they need it to in all their different development environments.

nicelhc13:
OK. Then I will prepare option "plans" for each application, but postpone the implementations. When everybody finishes their parts and you think it is proper timing, I will work on that later. Do you agree with this plan?

roshandathathri:
Sounds good.

nicelhc13:
[https://docs.google.com/spreadsheets/d/1EgHplvslmZGIkkJz0NdxMbCd60nDzTg_Laoa81j7zgo/edit?usp=sharing](https://docs.google.com/spreadsheets/d/1EgHplvslmZGIkkJz0NdxMbCd60nDzTg_Laoa81j7zgo/edit?usp=sharing)

The above doc contains consistent binary names and input graph specifying option lists.
Please note that the above doc is still in process.

For the naming, I followed two rules:
1) I only used abbreviations if I googled them and found the corresponding algorithms on the first page. 
2) I used dash ('-'), instead of under bar ('-') due to this article 
([https://www.mattcutts.com/blog/dashes-vs-underscores](https://www.mattcutts.com/blog/dashes-vs-underscores))

For the input graph formats, basically I exploited my backgrounds and descriptions.
If I don't know the format correctly, then I highlighted it through purple color.

I found that descriptions except for analytics cpu/gpu/dist looked insufficient.
[...]

PR #252, opened 2020-05-29
Input graph option (analytic/{cpu, distributed} and eda/cpu)

nicelhc13:
I didn't update lonestar/analytics/gpu, lonestar/mining/* and lonestar/scientific/* yet since they require more modifications.


nicelhc13:
My bad, please postpone to pull this request. I found that the distributed Galois has global-wise input graph checking option, which conflicts with my new added option.

nicelhc13:
Except pointstoanalysis (constraint file), analytic/gpu, and scientific/gpu , graph format checking is added to all the lonestar apps. Tests were done. Please give me any idea about handling *.cu files. 
Do you guys want to include llvm/clang headers and put graph format checking options? gpu files are not using llvm based options.

l-hoang:
I haven't looked at everything yet, but I think that the directed graph flag is overkill.

Even our symmetric graphs in Galois are technically directed graphs, so it adds no extra information that the user wouldn't already know. Also, there's nothing wrong with using a symmetric graph as input to an application like bfs.

l-hoang:
Second comment: distributed apps should not have a transposeGraph flag: CuSP automatically does an in-memory transpose as necessary depending on the partitioning policy used. Also, the flag conflicts with the graphTranspose flag which is used by CuSP to take in a transpose graph during partitioning for certain policies.

nicelhc13:
Directed graph option was suggested at  #161 . Please check and confirm this. I can remove those options. I will remove the transposed option soon. Thank you.

l-hoang:
"might take a directed one" is what I assume you're referring to. I'm not sure that's what he meant.

By default, all programs take a directed graph; a symmetric graph is just a directed graph with both directions. Therefore, in my opinion, there shouldn't be a directedGraph flag when this doesn't let the user know anything new about how to run a Galois program using the GR format.

The other suggestion is instead of having the symmetric/transpose flag located individually for every single benchmark, you can include it as a part of LonestarStart to (1) reduce code duplication and (2) make it uniform across all applications. The latter may be important for applications that may have different behavior depending on whether or not the user specifies transpose or symmetric.

One could argue the latter would add redundant command line options for apps that don't use them; the description for the CL arg should mention that it is optional for some graphs but required for others.

lonestar/analytics/distributed/tc/tc.cpp
```diff
@@ -136,6 +138,12 @@ int main(int argc, char** argv) {
   galois::DistMemSys G;
   DistBenchStart(argc, argv, name, desc, url);
 
+  if (!inputFileSymmetric) {
+    GALOIS_DIE("application requires a symmetric graph input;"
```
ddn0, line 142: Triangle counting requires ...

lonestar/analytics/cpu/ktruss/K-Truss.cpp
```diff
@@ -663,7 +655,7 @@ int main(int argc, char** argv) {
   LonestarStart(argc, argv, name, desc, url);
 
   if (!symmetricGraph) {
-    GALOIS_DIE("k-truss requires a symmetric graph input;"
+    GALOIS_DIE("K-Truss requires a symmetric graph input;"
```
ddn0, line 658: FWIW, the original intent was for "k-truss" to match the command run, which is why it isn't capitalized. Perhaps, a more general phrase, "This application requires ..." would be better in all cases?

ddn0:
And a general plea to squash/fixup some of these commits when this finally gets merged.

lonestar/analytics/cpu/ktruss/K-Truss.cpp
nicelhc13: That's a great idea. I didn't consider that. Then I am going to integrate all the error statements to "This application requires ~" or "[ERROR] This application requires ~."

The reason why I suggested the prefix "[ERROR]" is because it would be helpful for future wrappers which would exploit Galois as a back-end (e.g. Jupyter of the katana). This input-checking is not the only case that has attached the [ERROR]. All other errors should be attached to the prefix. Therefore, the front-end could parse/distinguish the errors, and basically could show the error statement (basically, copy-and-paste) to users.

What do you think about that?
I think it would be great to raise an issue if you agree with that.

l-hoang: GALOIS_DIE already prefixes the text with an error prefix IIRC.

l-hoang: As a side note, if you want to prefix some output text with an error prefix, galois::gError exists for that purpose.

nicelhc13: Thank you for letting me know, Loc. 
OK Then, 'This application' would be enough. I will work on that.

ddn0: @l-hoang took the words out of my mouth :D

ddn0: Regarding structured error messages: that's a big topic in itself. I think the what we should focus on here is just standardizing our own output just within Galois, in whatever form seems sensible regardless of downstream consumers. If it is well-structured here, then it is easy for folks downstream to do whatever processing they want. E.g., if we write all our error messages to stderr or use a specific logging format to distinguish debug, info, warning, fatal, then it will be easy for anyone downstream to pick up what they want and we can still be relatively decoupled.

libpangolin/include/pangolin/BfsMining/engine.h
```diff
@@ -1,10 +1,26 @@
 #include "galois/Galois.h"
 #include "pangolin/res_man.h"
 #include "pangolin/BfsMining/embedding_list.h"
+#include "llvm/Support/CommandLine.h"
+
+//! Flag that forces user to be aware that they should be passing in a
+//! symmetric graph.
+static cll::opt<bool> symmetricGraph(
+    "symmetricGraph",
+    cll::desc("Flag should be used to make user aware they should be passing a "
```
roshandathathri, line 10: This help message does not make sense. The flag is not for the user. It is for the system.

roshandathathri: Replace all such messages with either "The input graph is symmetric" or "Specify that the input graph is symmetric". 

On a different note, all help message should be consistent but they don't seem to be.

libpangolin/include/pangolin/BfsMining/engine.h
```diff
@@ -1,10 +1,26 @@
...
 int main(int argc, char** argv) {
   galois::SharedMemSys G;
   LonestarMineStart(argc, argv, name, desc, url);
+
+  if (!symmetricGraph) {
+    GALOIS_DIE("Mining apps requires a symmetric graph input;"
+               " please use the -symmetricGraph flag "
+               " to indicate the input is a symmetric graph");
+  }
+
```
l-hoang, line 23: You might want to mention that there's an assumption that the graph must be cleaned for mining applications as well.

That might require yet another "cleanGraph" flag.

libpangolin/include/pangolin/BfsMining/engine.h
nicelhc13: Noted! I will modify the command option to "Specify that the input graph is symmetric."

About the help message, yes I found the several inconsistent option descriptions, as you pointed out. Do you want me to work on that through this pull request? In my opinion, separating that would be more convenient for me, because they are not related to the symmetric graph option and this pull request already consists of a lot of modifications.

lonestar/liblonestar/src/BoilerPlate.cpp
```diff
@@ -34,6 +34,14 @@ llvm::cl::opt<std::string> statFile(
     llvm::cl::desc("ouput file to print stats to (default value empty)"),
     llvm::cl::init(""));
 
+//! Flag that forces user to be aware that they should be passing in a
+//! symmetric graph.
+llvm::cl::opt<bool> symmetricGraph(
+    "symmetricGraph",
+    llvm::cl::desc("Flag should be used to make user aware they should be passing a "
```
l-hoang, line 41: See Roshan's comment on the description for mining flag.

lonestar/mining/gpu/fsm/fsm_gpu.cpp
```diff
@@ -1,17 +1,35 @@
...
+
+  if (!symmetricGraph) {
+    GALOIS_DIE("Frequent-subgraph-mining requires a symmetric graph input;"
+               " please use the -symmetricGraph flag "
+               " to indicate the input is a symmetric graph");
+  }
+
   if (filetype != "adj") {
     std::cout << "Only support adj format for FSM\n";
+    std::cout << "Please add the -ft=adj flag\n";
```
l-hoang, line 32: May want to use galois::gError here.

lonestar/mining/gpu/kcl/kcl_gpu.cpp
```diff
@@ -1,18 +1,37 @@
 // Copyright 2019, University of Texas at Austin
 // Authors: Xuhao Chen <cxh@utexas.edu>
+#include "galois/Galois.h"
 #include "kcl.h"
 #include "lonestarmine.h"
+#include "llvm/Support/CommandLine.h"
```
l-hoang, line 6: Not sure what the interaction with llvm and GPU code is, though this is the CPP end so it's probably fine.

l-hoang:
Also, you'll have to fix the test errors: all the CI tests are failing.

This is most likely because the flags in question aren't being passed during CI.

libpangolin/include/pangolin/BfsMining/engine.h
roshandathathri: Yes, that should be a separate pull request.

Let this pull request be about graph options.

libpangolin/include/pangolin/BfsMining/engine.h
roshandathathri: Good point. However, I'm not sure whether "clean" term is well understood. The standard term is simple graph:
https://en.wikipedia.org/wiki/Multiple_edges
So let's use "-simpleGraph" and the help message should say "Specify that the graph is simple (has no multiple edges)".

roshandathathri:
Yes, CI tests must be updated before this can be merged.

libpangolin/include/pangolin/BfsMining/engine.h
l-hoang: Don't forget no self-loops as well.

nicelhc13: Noted

ddn0:
I didn't realize we were thinking that the extension was canonical 😅

If we go down this route, would there also be particular extensions for bipartite graphs, simple graphs, etc?

roshandathathri:
> For apps that accept only specific graph types, why don't we check file extensions instead of always requiring a specific flag?

Do you mean to use extensions for metis or aig graphs? That is fine with me but I would prefer the flag instead of the specific extension.
I'm not in favor of using extensions for flags like symmetric, transpose, or simple graphs.

roshandathathri:
> If we go down this route, would there also be particular extensions for bipartite graphs, simple graphs, etc?

I'm not in favor of this. For one, a graph can be both bipartite and simple. For another, it is adding a burden on the user to name their graphs (extensions) appropriately. I feel it's better to rely on them on passing the right flags.

gurbinder533:
I am also not in favor of using extensions. Certain applications can run on graphs with various extensions. We should be restricting graphs only for applications that can only run on those graphs or will fail.

insertinterestingnamehere:
Okay, it sounds like the file extension idea is not a popular idea. I still don't love having a flag that's required just to get the default behavior though, but we can do that if that's what people want. Are there any other alternatives?

In the long run a possible fix would be to make some kind of extension to the .gr format to allow space for graph properties. That's not doable for this release though.

ddn0:
> Okay, it sounds like the file extension idea is not popular. I still don't love having a flag that's required just to get the default behavior though, but we can do that if that's what people want. Are there any other alternatives?

I think from the perspective of a single application, you are right that having a flag to get the default behavior seems very weird.

From the perspective of a library of routines that are supposed to interoperate with each other, I can see the value of saying symmetric graphs are always referred to as "<filename> -symmetricGraph" regardless of application.

As implementers of the library, we are naturally biased towards the latter idea as well because it is simpler for us, but perhaps in this case, this might also match a user's perspective as well.

> In the long run a possible fix would be to make some kind of extension to the .gr format to allow space for graph properties. That's not doable for this release though.

I know some people who are working on extending the binary format of graphs....

insertinterestingnamehere:
Yah, I feel like these concerns could eventually be handled by graph format work elsewhere. I can kind-of see the rationale here after thinking about it some more. The idea is basically to make inputs named arguments on the command line instead of positional ones to reinforce the fact that some apps require specific kinds of inputs. Given that this idea has a lot of support here I'm good with merging it.

lonestar/analytics/cpu/pagerank/PageRank-pull.cpp
```diff
@@ -36,6 +36,13 @@ static cll::opt<Algo> algo("algo", cll::desc("Choose an algorithm:"),
                                        clEnumVal(Residual, "Residual")),
                            cll::init(Residual));
 
+//! Flag that forces user to be aware that they should be passing in a
+//! transposed graph.
+static cll::opt<bool> transposedGraph(
+    "transposedGraph",
```
l-hoang, line 42: This could be moved to the main LonestarStart.h header, but it's probably not used enough in other apps to need that since not many applications explicitly take a transpose graph.

It's fine as is.

lonestar/mining/gpu/fsm/fsm_gpu.cpp
```diff
@@ -1,17 +1,28 @@
...
 
 const char* name = "FSM";
 const char* desc = "Frequent subgraph mining in an undirected graph";
 const char* url  = 0;
 
 int main(int argc, char** argv) {
   LonestarMineStart(argc, argv, name, desc, url);
+
+  if (!simpleGraph) {
+    GALOIS_DIE("This application requires a simple graph input "
```
l-hoang, line 17: "no duplicate edges", not "no multiple edges"

lonestar/mining/gpu/kcl/kcl_gpu.cpp
```diff
@@ -1,18 +1,30 @@
...
 const char* name = "k-cliques";
 const char* desc = "Listing all k-cliques in an undirected graph";
 const char* url  = 0;
 
 int main(int argc, char** argv) {
   LonestarMineStart(argc, argv, name, desc, url);
+
+  if (!simpleGraph) {
+    GALOIS_DIE("This application requires a simple graph input "
+               " which has no multiple edges or self-loops;"
```
l-hoang, line 17: multiple edges -> duplicate edges

lonestar/mining/gpu/motif/motif_gpu.cpp
```diff
@@ -10,8 +12,17 @@ static int num_patterns[3] = {2, 6, 21};
 
 int main(int argc, char** argv) {
   LonestarMineStart(argc, argv, name, desc, url);
-  if (filetype != "gr") {
-    std::cout << "Only support gr format\n";
+
+  if (!simpleGraph) {
+    GALOIS_DIE("This application requires a simple graph input "
+               " which has no multiple edges or self-loops;"
```
l-hoang, line 18: multiple -> duplicate

lonestar/mining/gpu/motif/motif_gpu.cpp
```diff
@@ -10,8 +12,17 @@ static int num_patterns[3] = {2, 6, 21};
...
+               " please use the -simpleGraph flag "
+               " to indicate the input is a simple graph");
+  }
+
+  if (filetype != "gr") { 
+    galois::gError("Only support gr format\n"
```
l-hoang, line 24: "This application only supports...."

lonestar/mining/gpu/kcl/kcl_gpu.cpp
```diff
@@ -1,18 +1,30 @@
...
+               " please use the -simpleGraph flag "
+               " to indicate the input is a simple graph");
+  }
+
   if (filetype != "gr") {
-    std::cout << "Only support gr format\n";
+    galois::gError("Only support gr format\n"
```
l-hoang, line 23: "This application only supports..."

lonestar/mining/gpu/tc_mine/tc_mine_gpu.cpp
```diff
@@ -1,16 +1,27 @@
...
 const char* name = "Triangle counting";
 const char* desc = "Counting triangles in an undirected graph";
 const char* url  = 0;
 
 int main(int argc, char** argv) {
   LonestarMineStart(argc, argv, name, desc, url);
+
+  if (!simpleGraph) {
+    GALOIS_DIE("This application requires a simple graph input "
+               " which has no multiple edges or self-loops;"
```
l-hoang, line 17: multiple -> duplicate

lonestar/mining/include/lonestarmine.h
```diff
@@ -48,11 +48,24 @@ static cll::opt<std::string>
 static cll::opt<bool>
     verify("v", llvm::cl::desc("do verification step (default value false)"),
            llvm::cl::init(false));
+static cll::opt<bool>
+    simpleGraph("simpleGraph",
+                cll::desc("Specify that the input graph is simple (has no "
+                          "multiple edges or self-loops)"),
```
l-hoang, line 54: multiple -> duplicate

lonestar/mining/include/lonestarmine.h
```diff
@@ -48,11 +48,24 @@ static cll::opt<std::string>
...
+                cll::init(false));
 
 void LonestarMineStart(int argc, char** argv, const char* app, const char* desc,
                        const char* url) {
   // llvm::cl::SetVersionPrinter(LonestarPrintVersion);
   llvm::cl::ParseCommandLineOptions(argc, argv);
+
+  if (!simpleGraph) {
+    GALOIS_DIE("This application requires a simple graph input "
+               " which has no multiple edges or self-loops;"
```
l-hoang, line 64: multiple -> duplicate

roshandathathri:
@l-hoang @nicelhc13 The standard term is "no multiple edges". 
https://en.wikipedia.org/wiki/Multiple_edges
So do not change it to duplicate edges.

lonestar/mining/gpu/kcl/kcl_gpu.cpp
l-hoang: You missed this one. :P

roshandathathri:
> symSimpleGraph seems slightly clunky to me, but I think it's a class of graphs that is used in a lot of places, so it's fine.

I think it's clunky too. I would prefer independent flags that are simple and composable but I'm fine with a merged flag.

l-hoang:
I see it's been quashed; it will also make sense to resolve conflicts so CI can test the final, resolved version + we can review the merge here as well.

insertinterestingnamehere:
The merged version looks good to me. It needs a clang-format run again though.

lonestar/mining/include/lonestarmine.h
```diff
@@ -21,7 +21,7 @@ static cll::opt<unsigned>
             cll::init(1));
 static cll::opt<std::string>
     pattern_filename("p",
-                     cll::desc("<pattern graph filename: unsymmetrized graph>"),
+                     cll::desc("<pattern graph filename: symmetrized graph>"),
```
l-hoang, line 24: Are you sure the pattern file name is supposed to be symmetrized? I would make sure with someone familiar with the mining code.

nicelhc13: Yes. Xuhao confirmed this.

l-hoang: Cool, thanks.

ddn0:
We are so close! This is going to be great when this lands :balloon:

PR #253, opened 2020-05-29
b_lc_csr -> lc_csr_csc graph

l-hoang:
Renaming B_LC_CSR_Graph something more intuitive.


l-hoang:
Unrelated OMPI failure for a single dist test; merging now.

PR #255, opened 2020-05-30
dist readme update with performance, publications

l-hoang:
Updating the distributed readme with the sync/async options, performance considerations, and related publications.

### DIFF
--- a/lonestar/analytics/distributed/README.md
+++ b/lonestar/analytics/distributed/README.md
@@ -4,46 +4,42 @@ Overview of Distributed and Heterogeneous Systems in Galois
 This directory contains benchmarks that run using D-Galois and D-IrGL.
 
 D-Galois is distributed Galois built using the Gluon communication substrate.
-Similarly, D-IrGL is distributed IrGL built using Gluon. 
+Similarly, D-IrGL is distributed IrGL built using Gluon.
 Gluon is just the communication substrate: it is not a standalone system.
 
 Basic Compiling Through CMake (Distributed and Heterogeneous Galois)
 ================================================================================
 
 The dependencies for distributed Galois are exactly the same as shared-memory
-Galois except that it requires an MPI library (e.g. mpich2) to be on the 
+Galois except that it requires an MPI library (e.g. mpich2) to be on the
 system as well.
 
-To build distributed/heterogeneous Galois, certain CMake flags must be 
+To build distributed/heterogeneous Galois, certain CMake flags must be
 specified.
 
 For distributed Galois, i.e. D-Galois:
 
 `cmake ${GALOIS_ROOT} -DGALOIS_ENABLE_DIST=1`
 
-For heterogeneous (which is also distributed) Galois, i.e. D-IrGL:
+For distributed and heterogeneous Galois, i.e. D-IrGL:
 
 `cmake ${GALOIS_ROOT} -DGALOIS_ENABLE_DIST=1 -DGALOIS_ENABLE_GPU=1`
 
 Note that heterogeneous Galois requires CUDA 8.0 and above and a compiler
-that is compatible with the CUDA version that you use, and also note that
-enabling heterogeneous Galois also activates distributed Galois.
+that is compatible with the CUDA version that you use.
 
-Compiling with distributed/heterogeneous Galois will add the 'dist_apps' 
-directory to the build folder. Compiling with heterogeneous Galois enabled with 
-enable certain options in 'dist_apps'.
+Compiling with distributed Galois will add the `distributed` directory under
+`lonestar/analytics` to the build folder.
 
 Compiling Provided Apps
 ================================================================================
 
-Once CMake is successfully completed, you can build the provided apps with the 
-following command in the dist_apps directory.
+Once CMake is successfully completed, you can build the provided apps with the
+following command in lonestar/analytics/distributed directory.
 
 `make -j`
 
-You can compile a specific app by specifying it by name:
-
-`make -j bfs-push`
+You can compile specific apps by going their directories and running make.
 
 Running Provided Apps
 ================================================================================
@@ -54,16 +50,17 @@ command line option:
 `./bfs-push -help`
 
 Most of the provided graph applications take graphs in a .gr format, which
-is a Galois graph format that stores the graph in a CSR or CSC format. We 
+is a Galois graph format that stores the graph in a CSR or CSC format. We
 provide a graph converter tool under 'tools/graph-convert' that can take
 various graph formats and convert them to the Galois format.
 
 Running Provided Apps (Distributed Apps)
 ================================================================================
 
-First, note that if running multiple processes on a single machine, specifying
-`GALOIS_DO_NOT_BIND_THREADS=1` as an environment variable is crucial for 
-performance.
+First, note that if running multiple processes on a single machine (e.g.,
+single-host multi-GPU or multi-host multi-GPU where a process is spawned for
+each GPU), specifying `GALOIS_DO_NOT_BIND_THREADS=1` as an environment variable
+is crucial for performance.
 
 If using MPI, multiple processes split across multiple hosts can be specified
 with the following:
@@ -79,11 +76,19 @@ with the -help flag.
 Specifies the partitioning that you would like to use when splitting the graph
 among multiple hosts.
 
+`-exec=Sync,Async`
+
+Specifies synchronous communication (bulk-synchronous parallel where every host
+blocks for messages from other hosts at the end of a round of execution)
+or asynchronous communication (bulk-asynchronous parallel where a host does
+not have to block on messages from other hosts at the end of the round and
+may continue execution).
+
 `-graphTranspose`
 
-Specifies the transpose of the provided input graph. This is used to 
-create certain partitions of the graph (and is required for some of the 
-partitioning policies). It also makes 
+Specifies the transpose of the provided input graph. This is used to
+create certain partitions of the graph (and is required for some of the
+partitioning policies).
 
 `-runs`
 
@@ -96,7 +101,7 @@ Specify the file in which to output run statistics to.
 `-t`
 
 Number of threads to use on a single machine excluding a communication thread
-that is used by all of the provided distributed benchmarks. Note that 
+that is used by all of the provided distributed benchmarks. Note that
 GPUs only use 1 thread (excluding the communication thread).
 
 `-output` / `-outputLocation=<directory>`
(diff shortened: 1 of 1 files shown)


lonestar/analytics/distributed/README.md
```diff
@@ -54,15 +54,15 @@ command line option:
...
 Most of the provided graph applications take graphs in a .gr format, which
-is a Galois graph format that stores the graph in a CSR or CSC format. We 
+is a Galois graph format that stores the graph in a CSR or CSC format. We
 provide a graph converter tool under 'tools/graph-convert' that can take
 various graph formats and convert them to the Galois format.
 
 Running Provided Apps (Distributed Apps)
 ================================================================================
 
 First, note that if running multiple processes on a single machine, specifying
```
roshandathathri, line 64: It is not clear when multiple processes will be run. You can modify it to "machine (e.g., running on a single-host multi-GPU or multi-host multi-GPU machine)

lonestar/analytics/distributed/README.md
```diff
@@ -79,11 +79,19 @@ with the -help flag.
 Specifies the partitioning that you would like to use when splitting the graph
 among multiple hosts.
 
+`-exec=Sync,Async`
+
+Specifies synchronous communication (bulk-synchronous parallel where every host
+blocks for messages from other hosts at the end of a round of execution)
+or asynchronous communication (bulk-asynchronous parallel where a host does
+not have to block on messages from other hosts at the end of the round and
+may continue execution).
```
roshandathathri, line 88: Add a recommendation. "We recommend using Sync for 16 or fewer devices, and Async otherwise."

lonestar/analytics/distributed/README.md
```diff
@@ -114,22 +122,74 @@ Heterogeneous apps have additional command line parameters:
...
+"g" (GPUs). For example, if I have 2 machines with 8 GPUs each,
+but I want to run with 3 GPUs on each machine, I would use `-pset="ggg"`.
+Therefore, combined with `-num_nodes=2`, I would have a total of 6 units of
 execution: 3 GPUs on 2 machines for a total of 6. This creates a total of
 6 processes across the 2 machines (1 for each GPU).
 
-Also, it suffices to use only one "c" in pset to run on CPUs on your machines: 
-you can specify the amount of CPUs to use using the aforementioned thread 
+Also, it suffices to use only one "c" in pset to run on CPUs on your machines:
+you can specify the amount of CPUs to use using the aforementioned thread
```
roshandathathri, line 138: "the amount of cores/hyperthreads" instead of the amount of CPUs

lonestar/analytics/distributed/README.md
roshandathathri: Drop this. It's already handled below.

lonestar/analytics/distributed/README.md
```diff
@@ -114,22 +122,74 @@ Heterogeneous apps have additional command line parameters:
...
+Krishna Nandivada, Marc Snir, Keshav Pingali, “Gluon-Async: A Bulk-Asynchronous
+System for Distributed and Heterogeneous Graph Analytics,” Proceedings of the
+28th IEEE International Conference on Parallel Architectures and Compilation
+Techniques (PACT), September 2019.
+
+Vishwesh Jatala, Roshan Dathathri, Gurbinder Gill, Loc Hoang, V. Krishna
+Nandivada, Keshav Pingali, “A Study of Graph Analytics for Massive Datasets on
+Distributed GPUs,” Proceedings of the 34th IEEE International Parallel and
+Distributed Processing Symposium (IPDPS), May 2020.
+
```
roshandathathri, line 192: You should add our other papers as well. Abelian, MRBC, Phoenix, etc.

lonestar/analytics/distributed/README.md
```diff
@@ -114,22 +122,74 @@ Heterogeneous apps have additional command line parameters:
...
+Performance Considerations
+================================================================================
+
+* As mentioned above if running multiple processes on a single machine,
+  specifying `GALOIS_DO_NOT_BIND_THREADS=1` as an environment variable is
+  crucial for performance.
+
+* We have also observed that `GALOIS_DO_NOT_BIND_THREADS=1` to improve
+  performance in a distributed setting as well (multiple hosts each with its
+  own process).
```
roshandathathri, line 150: We should re-evaluate this. Maybe GALOIS_DO_NOT_BIND_MAIN_THREAD is a better option in this case.

roshandathathri: This is an orthogonal comment. It is not a requirement for this release or pull request.

lonestar/analytics/distributed/README.md
```diff
@@ -29,14 +29,14 @@ Note that heterogeneous Galois requires CUDA 8.0 and above and a compiler
 that is compatible with the CUDA version that you use, and also note that
```
roshandathathri, line 29: ENABLE_GPU does not ENABLE_DIST, so drop the part after "and ..."

lonestar/analytics/distributed/README.md
```diff
@@ -114,22 +122,74 @@ Heterogeneous apps have additional command line parameters:
...
+but I want to run with 3 GPUs on each machine, I would use `-pset="ggg"`.
+Therefore, combined with `-num_nodes=2`, I would have a total of 6 units of
 execution: 3 GPUs on 2 machines for a total of 6. This creates a total of
 6 processes across the 2 machines (1 for each GPU).
 
-Also, it suffices to use only one "c" in pset to run on CPUs on your machines: 
-you can specify the amount of CPUs to use using the aforementioned thread 
+Also, it suffices to use only one "c" in pset to run on CPUs on your machines:
+you can specify the amount of CPUs to use using the aforementioned thread
 option `-t`.
```
roshandathathri, line 139: After this, I would suggest adding a separate example section with a few examples - one each for distributed CPU-only, GPU-only and CPU+GPU runs.

lonestar/analytics/distributed/README.md
l-hoang: Abelian/MRBC sure. Phoenix, maybe not because it's not available anywhere in our code at the moment. Actually, even Abelian falls in the same bucket. Thoughts?

lonestar/analytics/distributed/README.md
l-hoang: Should I leave it in for now then or remove it out of precaution?

roshandathathri: Leave it as is.

lonestar/analytics/distributed/README.md
roshandathathri: Yes, it's just FYI.

l-hoang:
Updated; fixed some old references to dist_apps too (that was so long ago :P).

Take a look.

lonestar/analytics/distributed/README.md
```diff
@@ -4,46 +4,42 @@ Overview of Distributed and Heterogeneous Systems in Galois
...
-To build distributed/heterogeneous Galois, certain CMake flags must be 
+To build distributed/heterogeneous Galois, certain CMake flags must be
 specified.
 
 For distributed Galois, i.e. D-Galois:
 
 `cmake ${GALOIS_ROOT} -DGALOIS_ENABLE_DIST=1`
 
-For heterogeneous (which is also distributed) Galois, i.e. D-IrGL:
+For heterogeneous Galois, i.e. D-IrGL:
```
roshandathathri, line 24: This should be "For distributed and heterogeneous Galois". 

ENABLE_GPU does not enable heterogeneous galois, it also enables lonestar GPU. ENABLE_GPU with ENABLE_DIST enables heterogeneous Galois.

l-hoang: Good catch; relic of before the flag updates. Will fix.

lonestar/analytics/distributed/README.md
```diff
@@ -114,26 +119,118 @@ Heterogeneous apps have additional command line parameters:
...
-option `-t`.
+Also, it suffices to use only one "c" in pset to run on CPUs on your machines:
+you can specify the amount of cores/hyperthreads to use using the
+aforementioned thread option `-t`.
+
+Examples for Running Provided Apps
+================================================================================
+
+To run 3 processes all on a single machine, use the following:
+`mpirun -n=3 ./<app> <input-graph> -t=<num-threads-per-process>`
```
roshandathathri, line 142: This is incorrect. By default, num_nodes will be equal to n.

lonestar/analytics/distributed/README.md
```diff
@@ -114,26 +119,118 @@ Heterogeneous apps have additional command line parameters:
...
+
+To run on 3 CPUs on h1, h2, and h3, use the following:
+`mpirun -n=3 -hosts=h1,h2,h3 ./<app> <input-graph> -t=<num-threads>`
```
roshandathathri, line 145: I have not used the "hosts" option, so I don't know if it is correct.

lonestar/analytics/distributed/README.md
l-hoang: Right, if HETERO is on I'm assuming that's what it does. I'm too used to never building with hetero. 

Will fix.

lonestar/analytics/distributed/README.md
l-hoang: I used to use it a long time ago on our servers to debug things; it should work.

lonestar/analytics/distributed/README.md
```diff
@@ -114,26 +119,118 @@ Heterogeneous apps have additional command line parameters:
...
+
+To run on 3 GPUs on a single machine, use the following:
+`mpirun -n=3 ./<app> <input-graph> -t=<num-threads> -num_nodes=1 -pset="ggg"`
+
+To run on 4 GPUs on 2 machines h1 and h2 (each with 2 GPUs), use the following:
+`mpirun -n=4 -hosts=h1,h2 ./<app> <input-graph> -t=<num-threads> -num_nodes=2 -pset="gg"`
+Note that `mpirun -n=4` is 4 because there are a total of 4 execution units being used.
+
+To run on 1 CPU and 1 GPU each on 2 machines h1 and h2, use the following:
+`mpirun -n=4 -hosts=h1,h2 ./<app> <input-graph> -t=<num-threads> -num_nodes=2 -pset="cg"`
```
roshandathathri, line 155: In all cases, I would suggest adding -graphTranspose flag. I would also suggesting changing the partition for a few of them.

As these are examples, drop <...> and use the actual app or graph name.

lonestar/analytics/distributed/README.md
l-hoang: Actually, if HETERO isn't enabled IIRC num_nodes doesn't even show up. Should I add a note then?

roshandathathri: Thanks for catching that. Yes, it should only be applicable for multiple GPUs sharing a machine.

In the future, we can think of supporting multiple processes for multi-socket CPUs but this would require changing the binding of threads, so not the focus for now.

l-hoang:
Resolved; please check for correctness.

roshandathathri:
LGTM.

@l-hoang Thank you for handling this!

l-hoang:
@gurbinder533 You have until the CI tests finish to object to anything. :P

After that I'll merge.

PR #261, opened 2020-06-01
Graph in Pangolin -> PangolinGraph

l-hoang:
Graph is too common a keyword and doesn't tell a user where to
find the declaration. PangolinGraph makes this less of an issue.
Optimally everything in libpangolin should be in a pangolin
namespace; this can be resolved later.

This is in prep for Pangolin-Querying System integration being done elsewhere.


l-hoang:
Approved by Xuhao elsewhere: merging.

PR #262, opened 2020-06-02
Make non-distributed benchmarks output stats similar to distributed ones

ddn0:
This stats normalization will make it much easier for downstream systems to parse performance numbers. As it is now, there is not the same level of standardization of stats in the CPU applications as there is in the distributed ones. This makes parsing stats files very hairy.

Some of the input processing changes here overlap with #252 . I didn't realize there was another refactor of inputs was going on (great minds!).

I'd like to get the stats aspect in soonish, so let me know if I should reduce the scope of this change to just renaming the `StatTimer`s. Or what I should do to expedite things.

~I wasn't able to just put the `Input` stat processing in LonestarStart (c.f. DistBenchStart) because some applications don't take an input file at all (barneshut) or it may be optional (bipartite matching).~ edit: fixed by adding an argument to `LonestarStart`


lonestar/analytics/cpu/bipart/bipart.cpp
```diff
@@ -115,8 +114,8 @@ cell2) { cutsize++; cut_status = true; break;
  * Partitioning
  */
 void Partition(MetisGraph* metisGraph, unsigned coarsenTo, unsigned K) {
-  galois::StatTimer TM;
-  TM.start();
+  galois::StatTimer execTime("Timer_0");
+  execTime.start();
```
l-hoang, line 118: This is missing the corresponding execTime.stop().

lonestar/analytics/cpu/bipart/bipart.cpp
```diff
@@ -210,12 +208,15 @@ int hash(unsigned val) {
 
 int main(int argc, char** argv) {
   galois::SharedMemSys G;
-  LonestarStart(argc, argv, name, desc, url);
+  LonestarStart(argc, argv, name, desc, url, inputFile.c_str());
+
+  galois::StatTimer totalTime("TimerTotal");
+  totalTime.start();
```
l-hoang, line 214: Missing totalTime.stop()

lonestar/analytics/cpu/boruvka/Boruvka.cpp
```diff
@@ -423,7 +422,10 @@ void run() {
 
 int main(int argc, char** argv) {
   galois::SharedMemSys G;
-  LonestarStart(argc, argv, name, desc, url);
+  LonestarStart(argc, argv, name, desc, url, inputFile.c_str());
+
+  galois::StatTimer totalTime("TimerTotal");
+  totalTime.start();
```
l-hoang, line 428: Missing stop.

Since I'm seeing this somewhat consistently, I have to ask: is this being done on purpose? i.e. you're just letting the destructor take care of things. Does the destructor save the time when it gets destroyed, though?

lonestar/analytics/cpu/clustering/leidenClustering.cpp
```diff
@@ -316,20 +315,21 @@ void runMultiPhaseLouvainAlgorithm(Graph& graph, uint64_t min_graph_size,
 
 int main(int argc, char** argv) {
   galois::SharedMemSys G;
-  LonestarStart(argc, argv, name, desc, url);
+  LonestarStart(argc, argv, name, desc, url, inputFile.c_str());
 
-  Graph graph, graph_next;
-  Graph* graph_curr;
+  galois::StatTimer totalTime("TimerTotal");
+  totalTime.start();
```
l-hoang, line 321: Stop missing.

See other comment for whether or not this is being done on purpose :p

I'm just marking them here in case it wasn't. (though I might miss some of them due to human error)

lonestar/analytics/cpu/clustering/louvainClustering.cpp
```diff
@@ -792,20 +790,21 @@ void runMultiPhaseLouvainAlgorithm(Graph& graph, uint32_t min_graph_size,
 
 int main(int argc, char** argv) {
   galois::SharedMemSys G;
-  LonestarStart(argc, argv, name, desc, url);
+  LonestarStart(argc, argv, name, desc, url, inputFile.c_str());
 
-  Graph graph, graph_next;
-  Graph* graph_curr;
+  galois::StatTimer totalTime("TimerTotal");
+  totalTime.start();
```
l-hoang, line 796: stop missing.

lonestar/analytics/cpu/connectedcomponents/ConnectedComponents.cpp
```diff
@@ -1206,10 +1210,17 @@ void run() {
 
 int main(int argc, char** argv) {
   galois::SharedMemSys G;
-  LonestarStart(argc, argv, name, desc, url);
+  LonestarStart(argc, argv, name, desc, nullptr, inputFile.c_str());
+
+  galois::StatTimer totalTime("TimerTotal");
+  totalTime.start();
```
l-hoang, line 1216: missing stop

lonestar/analytics/cpu/gmetis/GMetis.cpp
```diff
@@ -181,13 +173,16 @@ typedef galois::substrate::PerThreadStorage<std::map<GNode, uint64_t>>
 
 int main(int argc, char** argv) {
   galois::SharedMemSys G;
-  LonestarStart(argc, argv, name, desc, url);
+  LonestarStart(argc, argv, name, desc, url, inputFile.c_str());
+
+  galois::StatTimer totalTime("TimerTotal");
+  totalTime.start();
```
l-hoang, line 179: missing stop.

Starting to think this is on purpose...

lonestar/analytics/cpu/independentset/IndependentSet.cpp
```diff
@@ -696,7 +702,16 @@ void run() {
 
 int main(int argc, char** argv) {
   galois::SharedMemSys G;
-  LonestarStart(argc, argv, name, desc, url);
+  LonestarStart(argc, argv, name, desc, url, inputFile.c_str());
+
+  galois::StatTimer totalTime("TimerTotal");
+  totalTime.start();
```
l-hoang, line 708: missing stop

lonestar/analytics/cpu/kcore/kcore.cpp
```diff
@@ -232,7 +229,10 @@ void kCoreSanity(Graph& graph) {
 
 int main(int argc, char** argv) {
   galois::SharedMemSys G;
-  LonestarStart(argc, argv, name, desc, url);
+  LonestarStart(argc, argv, name, desc, nullptr, inputFile.c_str());
+
+  galois::StatTimer totalTime("TimerTotal");
+  totalTime.start();
```
l-hoang, line 235: Missing stop.

Going to stop marking them from this point onward because I'm 99% percent sure this is on purpose at this point (I just checked the destructor and it does indeed call stop).

One thing that I do want to comment about though is that this may be confusing to a code reader who isn't aware that the destructor for a StatTimer calls stop. One could argue that it's something obvious, but leaving things up to assumption doesn't sit well with me.

For the record, I'm fine with the way it currently is too; I simply prefer to be as explicit as possible in code to avoid confusion to new code readers.

lonestar/scientific/cpu/barneshut/Barneshut.cpp
```diff
@@ -563,7 +563,7 @@ void run(Bodies& bodies, BodyPtrs& pBodies, size_t nbodies) {
 
 int main(int argc, char** argv) {
   galois::SharedMemSys G;
-  LonestarStart(argc, argv, name, desc, url);
+  LonestarStart(argc, argv, name, desc, url, nullptr);
 
   std::cout << config << "\n";
   std::cout << nbodies << " bodies, " << ntimesteps << " time steps\n";
```
l-hoang, line 569: This app is missing the execTimer("Timer_0") and TimerTotal timers.

lonestar/analytics/cpu/pagerank/PageRank-pull.cpp
```diff
@@ -258,27 +259,28 @@ void prResidual(Graph& graph) {
 
   initNodeDataResidual(graph, delta, residual);
   computeOutDeg(graph);
-  galois::StatTimer prTimer;
-  prTimer.start();
+
+  galois::StatTimer execTime("Timer_0");
```
l-hoang, line 263: Just a comment, but this is one of those unfortunate cases where you have Timer_0 in more than one place due to how the code is laid out.

PR #265, opened 2020-06-03
Change bipart test back to the ibm01 input to avoid running out of memory in CI.

insertinterestingnamehere:


insertinterestingnamehere:
Failures are unrelated. Merging.

PR #267, opened 2020-06-03
Fix evaluation order in BoilerPlate input option

ddn0:
Command line option objects are populated during LonestarStart call so
we cannot pass their values to LonestarStart directly. Pass a pointer
instead.

I also considered using std::optional<> instead of a pointer here, but
std::optional for reference types is not in the C++17 standard.


insertinterestingnamehere:
We have a major incoming reorganization from @nicelhc13 that will likely conflict with this. This is a great improvement though. @nicelhc13, depending on how far along you are with the reorganization, we may want to merge this first.

ddn0:
Yup; I'll let @nicelhc13 decide how to incorporate these changes.

insertinterestingnamehere:
Okay, I e-mailed him directly. He's still working on https://github.com/IntelligentSoftwareSystems/Galois/pull/252. The CI builds should be done in a few minutes so we can just merge this as soon as they finish.

PR #268, opened 2020-06-03
Report error when k argument is invalid

ddn0:
For user-controllable parameters, it makes more sense to fail more
visibly when values are invalid rather than using an assertion. The
latter will only report an issue in a debug build.

There seems to be some requirement to not depend on libgalois in the mining code; otherwise, I might have used `GALOIS_ASSERT` and `gError` rather than `printf`. Also there seems to be some confusing naming where the AppMiner parameter is called `ms` (and there is a command line parameter `ms`) while the command line option is `k`. I'm not too familiar with this part of the codebase, so I left things as is.

### DIFF
--- a/lonestar/mining/cpu/fsm/fsm.cpp
+++ b/lonestar/mining/cpu/fsm/fsm.cpp
@@ -14,7 +14,10 @@ class AppMiner : public EdgeMiner<LabeledElement, EdgeEmbedding, MyAPI, true> {
 public:
   AppMiner(unsigned ms, int nt)
       : EdgeMiner<LabeledElement, EdgeEmbedding, MyAPI, true>(ms, nt) {
-    assert(ms > 1);
+    if (ms <= 1) {
+      printf("ERROR: command line argument k must be 2 or greater\n");
+      exit(1);
+    }
     if (filetype == "gr") {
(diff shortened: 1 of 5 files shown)


ddn0:
cc @gurbinder533

l-hoang:
I'm pretty sure it's intentional.

The problem with the code at the moment is that NVCC hates Galois.h, so when you compile the GPU end, things will break spectacularly. This is more of an issue with the way the code happens to be organized than anything else. (i.e., code needs to be refactored at some point).

ddn0:
Cool; I'll merge this in now; it maintains consistency with the surrounding code so it is no worse than before. If we think rationalize the structure of the mining apps is important, someone can file a ticket :D

roshandathathri:
> The problem with the code at the moment is that NVCC hates Galois.h, so when you compile the GPU end, things will break spectacularly.

Yes, but we structure distributed analytics apps without facing such issues. It should be possible to do the same for mining apps.

> it maintains consistency with the surrounding code so it is no worse than before

Agreed.

> I wasn't sure if this is intentional

Don't think it is intentional.

l-hoang:
It's intentional in the sense that the code writer avoided using those functions else the GPU build would break. Yes, it should be refactored such that this isn't an issue.

PR #269, opened 2020-06-04
Fixing LCI warnings

roshandathathri:


roshandathathri:
> Needs a clang-format run

That's strange. I had run check-format on the files I changed.

l-hoang:
clang-format on non-server machines (i.e. fusion, velocity, etc.) is a lower version than the one being used to cross check. I usually have to run on it on peltier, for instance.

roshandathathri:
Thanks, Loc! Fixed it and pushed.

l-hoang:
Something strange is going on with the arch build: been seeing that same error in other places as well.

Otherwise everything else passes so I'll just go ahead and merge it.

Issue #185, opened 2020-05-06
Compiler Bug in Gcc 7.2 C++17 Support


insertinterestingnamehere:
Currently, when building with gcc 7.2 or 7.1 we get a bunch of errors of the form `error: uninitialized variable ‘thing’ in ‘constexpr’ function` in many of the Lonestar apps. As far as I can tell, this is an example of https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80935.

Given that there's not an obvious workaround other than dropping C++17 or dropping our use of lambdas for parallel loops in the Lonestar apps, I propose we require gcc 7.3 or higher instead of just gcc 7. Usually I don't worry much about requiring newer minor release versions, however on our development machines, gcc 7.2 is the only gcc version in the gcc 7 series we currently have set up. It doesn't look like this will cause any issues with any of the major Linux distros though. In addition, it's very likely that we will be updating the OS setup on our development machines soon anyway.

Is requiring gcc 7.3 or later okay with everyone?

roshandathathri:
I wouldn't mind but Stampede only has gcc 7.1 (with MPI and boost support). I wouldn't want to make it hard for us to use Galois on Stampede or any other cluster we use. 

What are our other options?

l-hoang:
Stampede2 also has gcc 7.3 and 9.1, but 9.1 if I recall had issues because it doesn't have a compatible boost version that we can use.

roshandathathri:
I was talking about Stampede2 only. Neither gcc 7.3 nor gcc 9.1 have compatible MPI or boost modules.

insertinterestingnamehere:
I'm not sure there are other good options. I can hack on it a bit more and try to find a workaround, but, looking at the bug report, a reliable workaround seems unlikely. Note, however, that this is only in the Lonestar apps, not the core library.

insertinterestingnamehere:
Note: the gcc 7.1 binaries they have should work fine without any modification with binaries built with the gcc 7.3 compiler. I don't know if they've configured their modules to be aware of that fact though.

roshandathathri:
> Note: the gcc 7.1 binaries they have should work fine without any modification with binaries built with the gcc 7.3 compiler. I don't know if they've configured their modules to be aware of that fact though.

No, they haven't. The modules are not available for later versions as of now.

insertinterestingnamehere:
[...]

PR #275, opened 2020-06-05
Work around bug in gcc 7.1 and 7.2

insertinterestingnamehere:
Fixes https://github.com/IntelligentSoftwareSystems/Galois/issues/185. It turned out that just zero initializing the variables it was complaining about made the error go away.

### DIFF
--- a/lonestar/eda/cpu/SPRoute/maze_finegrain.h
+++ b/lonestar/eda/cpu/SPRoute/maze_finegrain.h
@@ -1340,7 +1340,7 @@ void mazeRouteMSMD_finegrain_spinlock(int iter, int expand, float costHeight,
                   }
                   // curr_d1 = d1_push;
 
-                  int preX, preY;
+                  int preX = curX, preY = curY;
                   if (curr_d1 != 0) {
                     if (HV[curY][curX]) {
                       preX = parentX1[curY][curX];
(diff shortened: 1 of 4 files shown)


roshandathathri:
Great. Thanks, @insertinterestingnamehere !

roshandathathri:
LGTM

lonestar/eda/cpu/SPRoute/maze_finegrain.h
```diff
@@ -2338,7 +2335,7 @@ void mazeRouteMSMD_finegrain_doall(int iter, int expand, float costHeight,
                         galois::runtime::acquire(&data[grid],
                     galois::MethodFlag::WRITE);*/
 
-                    int preX, preY;
+                    int preX = curX, preY = curY;
```
l-hoang, line 2338: Already merged, but comment here:

Before the change,  I believe it was possible for preX and preY to remain uninitialized even after the following if condition if (1) curr_d1 != 0 AND (2) HV[curY][curX] was 0.

Does this code rely on preX/preY being uninitialized anywhere @Michael-JY-He?

insertinterestingnamehere: Relying on an uninitialized value is undefined behavior, so I hope not. In the case of `preX` and `preY` though, it's not possible. The else clause below used to cover that. That said, I wasn't able to confirm that `tmp`, `tmp_cost`, `tmp_grid`, `tmpX`, and `tmpY` are always initialized. It would be nice to know.

l-hoang: I also hope there was no reliance on that.

In the old code it was possible for it to be uninitialized if you entered the first if but not the second if (thereby ignoring the else clause completely).

Michael-JY-He: The new code looks good to me. But I still don't understand why the old code is uninitialized? Everybranch will initialize preX and preY right?

l-hoang: There are 3 possibilites:

1) curr_d1 != 0 AND HV[curY][curX] != 0
2) curr_d1 != 0 AND HV[curY][curX] **== 0**
3) curr_d1 == 0

(1) and (3) are fine. (2) is where preX and preY will not be initialized if I understand correctly.

There may be an invariant I'm not aware of that prevents (2) from happening.

l-hoang: Wait, never mind, the code preview was hiding the "else" that handles case (2) from me. My bad.

PR #277, opened 2020-06-05
Adding GALOIS_CUDA_CAPABILITY cmake option and fixing output for distributed apps

### LINKED ISSUE #273: Make GALOIS_CUDA_CAPABILITY cmake option
From @roshandathathri:
> Merge `GALOIS_ENABLE_GPU` and `CUDA_CAPABILITY` into a single GALOIS_CUDA_CAPABILITY with a default of none. Although it is not a simple "enable" flag, I think it makes it more explicit that the user must set the compute capability.

roshandathathri:
Closes #273


CMakeLists.txt
```diff
@@ -180,14 +179,18 @@ include(CheckEndian)
...
+elseif(EXISTS /net/ohm/export/iss)
   set(BASEINPUT /net/ohm/export/iss/inputs)
-  MESSAGE(STATUS "Using graph location /net/ohm/export/iss/inputs")
+  MESSAGE(STATUS "Using graph input location /net/ohm/export/iss/inputs")
+  set(BASEOUTPUT /net/ohm/export/iss/dist-outputs)
+  MESSAGE(STATUS "Using graph output location /net/ohm/export/iss/dist-outputs")
 else()
   set(BASEINPUT "${PROJECT_BINARY_DIR}/inputs")
-  message(STATUS "Use 'make input' to download inputs in the build directory")
+  set(BASEOUTPUT "${PROJECT_BINARY_DIR}/inputs")
```
ddn0, line 192: should this be `${PROJECT_BINARY_DIR}/outputs`?

lonestar/CMakeLists.txt
```diff
@@ -82,6 +74,7 @@ if(GALOIS_ENABLE_DIST)
 
   set(RESULT_CHECKER ${PROJECT_SOURCE_DIR}/scripts/result_checker.py)
   cmake_host_system_information(RESULT HOSTNAME QUERY HOSTNAME)
+  file(MAKE_DIRECTORY "${CMAKE_BINARY_DIR}/output")
```
ddn0, line 77: `${BASEOUTPUT}`?

lonestar/analytics/distributed/README.md
```diff
@@ -23,7 +23,12 @@ For distributed Galois, i.e. D-Galois:
 
 For distributed and heterogeneous Galois, i.e. D-IrGL:
 
-`cmake ${GALOIS_ROOT} -DGALOIS_ENABLE_DIST=1 -DGALOIS_ENABLE_GPU=1`
+`cmake ${GALOIS_ROOT} -DGALOIS_ENABLE_DIST=1 -DGALOIS_CUDA_CAPABILITY=<insert CUDA capability here>`
+
+The CUDA capability should be one that your GPU supports. For example, if I
```
ddn0, line 28: I don't know what the doc style is for voice, but `I` here seems out of place to me, since that presumably refers to the author not the reader. "You" instead?

CMakeLists.txt
```diff
@@ -203,9 +206,10 @@ if (GALOIS_ENABLE_DIST)
   add_subdirectory(libcusp)
   add_subdirectory(libgluon)
 endif()
+string(COMPARE NOTEQUAL "${GALOIS_CUDA_CAPABILITY}" "" GALOIS_ENABLE_GPU)
 if (GALOIS_ENABLE_GPU)
   enable_language(CUDA)
-  string(REPLACE "." "" GENCODES ${CUDA_CAPABILITY})
+  string(REPLACE "." "" GENCODES ${GALOIS_CUDA_CAPABILITY})
   string(REPLACE "," ";" GENCODES ${GENCODES})
```
ddn0, line 213: I wasn't around when `${CUDA_CAPABILITY}` was introduced but was there a reason that we used a comma-separated `,` list rather than semi-colon `;`? The latter is more commonly used to separate list items in CMake and we probably could get rid of the `string(REPLACE "," ";" ...)` below.

lonestar/analytics/distributed/README.md
```diff
@@ -140,8 +145,9 @@ Examples for Running Provided Apps
 
 To run 3 processes all on a single machine, use the following:
 `GALOIS_DO_NOT_BIND_THREADS=1 mpirun -n=3 ./bfs_push rmat15.gr -graphTranspose=rmat15.tgr -t=4 -num_nodes=1 -partition=oec`
-Note that `-num_nodes=1` is not correct if `ENABLE_GPU` is not set (it does not
-appear as an option in that case).
+Note that `-num_nodes=1` is not correct if heterogeneous execution is not
```
ddn0, line 148: The wording here seems a little confusing to me (though I'm not super familiar with the interface). Would something like: "Note: when heterogeneous execution is enabled, the setting `-num_nodes=1` is invalid and will also not appear as an option." mean the same thing here and be a little clearer? 

Optionally afterwards, "Heterogeneous execution is enabled by setting the `GALOIS_CUDA_CAPABILITY` build option."

l-hoang: Note: when heterogeneous execution is **not** enabled, `-num_nodes=1` is invalid and will not appear as an option.

Yes, much clearer. My bad for the poor wording.

lonestar/analytics/distributed/README.md
l-hoang: "You" does make more sense.

CMakeLists.txt
roshandathathri: No, it's intentional. For now, I'm keeping the inputs and outputs directory the same (except those hard-coded paths). I'll leave to Ian to change that if he changes "make input" to make them different folders.

lonestar/CMakeLists.txt
roshandathathri: No, BASEOUTPUT has the reference outputs. This directory is for the outputs generated for the test runs (i.e., the outputLocation).

CMakeLists.txt
roshandathathri: I'm fine with either. I had used comma-separated because I thought it was more commonly used. However, as we're using semi-colon list for other cmake options, I'll change this option to semi-colon as well to be consistent.

roshandathathri:
Arch failure seems unrelated. @insertinterestingnamehere any idea why?

l-hoang:
I've been seeing that error occur non-deterministically on arch FWIW.

insertinterestingnamehere:
Yes, the Arch build failure is unrelated. I'm not sure what's causing it yet.

roshandathathri:
Thanks, Loc and Ian. I'm merging this PR then.

PR #283, opened 2020-06-07
Rename lonestar directories and binaries

nicelhc13:
Renamed based on [the doc](https://docs.google.com/spreadsheets/d/1EgHplvslmZGIkkJz0NdxMbCd60nDzTg_Laoa81j7zgo/edit#gid=0).
'Make test' were successfully done.
Please double check if there is any mistake.


lonestar/CMakeLists.txt
```diff
@@ -23,6 +23,36 @@ function(add_test_scale type app)
...
+    add_executable(${target_name} ${name}.cu support.cu)
+  else()
+    add_executable(${target_name} ${name}.cu)
+  endif()
+  install(TARGETS ${target_name} DESTINATION "${CMAKE_INSTALL_BINDIR}" EXCLUDE_FROM_ALL)
+  if(GALOIS_ENABLE_GPU AND NOT ${X_NO_GPU})
+    target_compile_options(${target_name} PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:-w>)
+    target_link_libraries(${target_name} Galois::gpu)
+    set_property(TARGET ${target_name} PROPERTY CUDA_STANDARD 14)
+    set_property(TARGET ${target_name} PROPERTY CUDA_SEPARABLE_COMPILATION ON)
```
ddn0, line 51: quick question: is setting `CUDA_SEPARABLE_COMPILATION=on` actually required?

lonestar/CMakeLists.txt
```diff
@@ -23,6 +23,36 @@ function(add_test_scale type app)
   endforeach()
 endfunction(add_test_scale)
 
+function(app_cpu PREFIX name bin_name)
+  string(CONCAT target_name ${PREFIX} ${bin_name})
```
ddn0, line 27: FWIW, the original idea behind getting rid of the CMake `app` function was to reduce the chance of introducing inadvertent coupling through a shared `app` function.

Also, it turns out most of the build configuration for a lonestar app is exactly the same as building any application in CMake so hiding this behind a function might be more confusing for people who just want to use Galois libraries in their own projects.

Anyway, I can see the value of reducing duplication in the CMake files but I wanted to raise the fact that I went down this path before and wanted to go the other way because `app` was making things more brittle.

lonestar/analytics/CMakeLists.txt
```diff
@@ -1,9 +1,29 @@
+function(app_analy_cpu name bin_name)
```
ddn0, line 1: It seems like a lot of extra complexity to introduce these types of functions for the sole purpose of prefixing targets when the directory structure in the build tree is something only relevant to developers (and subject to future change as we reorganize things) while most users would be using the flattened structure in the installed bindir.

lonestar/CMakeLists.txt
```diff
@@ -23,6 +23,36 @@ function(add_test_scale type app)
...
+  add_executable(${target_name} ${name}.cpp)
+  add_dependencies(apps ${target_name})
+  target_link_libraries(${target_name} PRIVATE Galois::shmem lonestar)
+  install(TARGETS ${target_name} DESTINATION "${CMAKE_INSTALL_BINDIR}" COMPONENT apps EXCLUDE_FROM_ALL)
+  set_target_properties(${target_name} PROPERTIES OUTPUT_NAME ${bin_name})
```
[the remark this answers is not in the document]
nicelhc13: I see. The first motivation of this work is that we want to create multiple binaries having the same name within the same CMake project. Now, it seems like we cannot avoid prefix, then.
For example, the analytic cpu tc, the analytic dist tc, and the mining cpu tc would be 'analy-cpu-triangle-counting', 'analy-dist-triangle-counting', 'mining-cpu-triangle-counting.'

What do you think about this? Do you have any advice on this? Thank you.

lonestar/CMakeLists.txt
nicelhc13: Sure. I also agreed with you, but didn't know how to balance the reducing-codes and the inadvertent code couplings. Also as you can see, some scientific apps and the gpu apps required separate codes from the functions, which looked very messy.

For this issue, I would like to hear opinions from distributed and GPU Galois teams since they maintained the style coupling the codes through the functions.
If all agree with that,  I could decouple the codes from app_cpu(), app_dist(), and app_gpu(). 
Please correct if I misunderstood.

lonestar/CMakeLists.txt
vishweshjatala: Some of the LonestarGPU applications call kernels within the kernel. Enabling this parameter resolves the compilation issues for such applications.

roshandathathri: Yes, it's required (leads to errors otherwise) for distributed GPU code as well.

lonestar/CMakeLists.txt
roshandathathri: This is debatable. It depends on the app or set of apps. The question is whether we want to be consistent. In any case, it should be a separate issue/commit/pull-request.

lonestar/CMakeLists.txt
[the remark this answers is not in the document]
roshandathathri: I would a prefix a unique target/binary name in the entire project. It will keep things simple. 

Instead of a prefix, I would suggest a suffix. For example, triangle-counting-cpu, triangle-counting-gpu, triangle-counting-dist. I would also suggest to use these suffixes consistently for all apps. 
I would suggest to NOT use suffix for "type" of the app unless required. In the other words, do not use "analytics" and only use "mining" when you have to (that is, for triangle-counting-mining-cpu and triangle-counting-mining-gpu).

lonestar/analytics/CMakeLists.txt
```diff
@@ -1,9 +1,29 @@
...
+endfunction()
+
 add_subdirectory(cpu)
 
 if(GALOIS_ENABLE_DIST)
   add_subdirectory(distributed)
 endif()
 
-if(GALOIS_ENABLE_GPU)
+if(GALOIS_CUDA_CAPABILITY)
```
roshandathathri, line 27: Use GALOIS_ENABLE_GPU here. 

GALOIS_ENABLE_GPU CMake variable still exists. It is not set by the user, but derived using GALOIS_CUDA_CAPABILITY.

roshandathathri:
@nicelhc13 please run "make test" on some UT machine that has GPUs. I want to make sure that distributed (GPU) apps run and their output is verified.

nicelhc13:
> @nicelhc13 please run "make test" on some UT machine that has GPUs. I want to make sure that distributed (GPU) apps run and their output is verified.

Hi Roshan, I already did the 'make test' on the tuxedo machine.
analytics GPU apps do not have 'make test' so I raised an issue yesterday.

vishweshjatala:
I will fix the missing test cases for lonestarGPU apps that I own.

lonestar/CMakeLists.txt
insertinterestingnamehere: I agree. Let's not introduce these extra functions. The duplication is small and it makes the CMake configuration files much harder to read for someone unfamiliar with the project.

lonestar/analytics/CMakeLists.txt
insertinterestingnamehere: Do we expect users to use installed versions of the apps?

lonestar/analytics/cpu/k-truss/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+app_analy_cpu(K-Truss k-truss)
+add_custom_command(
```
insertinterestingnamehere, line 2: Does anyone know why we have bmk-related code in the CMakeLists.txt for k-truss? Is this deliberate?

lonestar/mining/cpu/frequent-subgraph-mining/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+app_mining_cpu(fsm frequent-subgraph-mining)
+
+add_test_mine(small1 frequent-subgraph-mining -symmetricGraph -simpleGraph "${BASEINPUT}/Mining/citeseer.sadj" "-ft=adj")
```
insertinterestingnamehere, line 3: This was commented out before due to https://github.com/IntelligentSoftwareSystems/Galois/issues/143. Please leave this commented out until I fix that so we can just get these fixes through more quickly.

lonestar/mining/cpu/motif-counting/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+app_mining_cpu(motif motif-counting)
+
+add_test_mine(small1 motif-counting -symmetricGraph -simpleGraph "${BASEINPUT}/Mining/citeseer.csgr")
```
insertinterestingnamehere, line 3: Same as above. Just leave this commented out to avoid depending on https://github.com/IntelligentSoftwareSystems/Galois/issues/143.

lonestar/mining/cpu/triangle-counting/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+app_mining_cpu(tc_mine triangle-counting)
+
+add_test_mine(small1 triangle-counting -symmetricGraph -simpleGraph "${BASEINPUT}/Mining/citeseer.csgr")
```
insertinterestingnamehere, line 3: Comment this out for now.

lonestar/analytics/CMakeLists.txt
yishanlu: I don't think so. The users are mostly interested in using Galois as a library if they have their own applications to be Galoised.

lonestar/analytics/cpu/k-truss/CMakeLists.txt
yishanlu: The bmk related code in CMake here was introduced by Sree to run through configurations for collecting performance numbrers. I think we should remove this for the release.

roshandathathri: This was added by Sree. We should see if we can make it consistent with the other lonestar apps in a separate issue/commit.

nicelhc13: Noted. I will remove it.

lonestar/mining/cpu/frequent-subgraph-mining/CMakeLists.txt
nicelhc13: Oh  I see. I removed comments of some mining tests. I am going to revert them back.

lonestar/CMakeLists.txt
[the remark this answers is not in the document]
insertinterestingnamehere: I favor this too, though it contradicts what we discussed on Thursday. I was under the impression that we wanted to make the executable naming the same for the different variants to avoid inconveniencing the people developing the distributed/GPU versions. If this sounds better to everyone, let's do it this way. It will likely be easier to manage in the long run. Unfortunately it'll require quite more changes from @nicelhc13 here before we merge though.

One small nitpick with the naming. Is the cpu suffix really needed? Could we just do `triangle-counting` and `triangle-counting-dist`?

lonestar/analytics/cpu/k-truss/CMakeLists.txt
roshandathathri: Please do not do this in this commit or pull-request. Renaming is a big change - get that out of way. Anything else should be consistent with what was there before the renaming.

lonestar/CMakeLists.txt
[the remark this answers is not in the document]
roshandathathri: There's triangle-counting-gpu as well. We can say the default (without suffix) is "cpu" but I think that add some overhead for the user. It's better to be explicit for the architecture in my opinion. Analytics/mining - there it's fine to omit where possible.

lonestar/analytics/CMakeLists.txt
l-hoang: Installing the apps seems very unlikely to me; most people probably don't need a graph analytics routine they can call on the fly.

lonestar/CMakeLists.txt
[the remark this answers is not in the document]
insertinterestingnamehere: Leaving the cpu suffix is also fine with me. With or without it, this is still by far the best naming scheme we've come up with thus far.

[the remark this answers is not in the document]
nicelhc13: I have one more question! To sum up, our rule is like below, to my best understanding.
1. Specify architecture type through an additional suffix (e.g. *-cpu, *-dist, *-gpu). In this case, dist attached apps could use either cpu or gpu, or both of them.

2. Omit the application types as many as possible. (It is a confusing part for me)
Based on the rule which shall omit 'analytic' suffix, do you also want to remove a suffix if the app is unique across the lonestar? For example, let's assume that lonestar/scientific/cpu only has delaunarytriangulation. In this case, that would be the 'delaunarytriangulation-cpu', not the 'delaunarytriangulation-scientific-cpu.'

Am I correct?

lonestar/CMakeLists.txt
nicelhc13: So,
1) Do you guys want to handle this issue on the separate pull or issue?
If 1) is yes, Then 2) Do you guys want to keep the shared function for now? or want to separate them?

Please confirm those two questions. Then I could restart this request.

roshandathathri: If you can, keep it separate. If you have already done something, you don't need to undo it. I would suggest merging this as soon as possible. So try to minimize the changes you do.

lonestar/analytics/CMakeLists.txt
ddn0: Insofar as most users of this repo are other developers, I agree that using the build tree is a convenient way to refer to executables, but, if I were instead someone interested in installing a suite of graph analytics applications for myself and others, I would naturally assume that the installed executables are the primary artifact. (In the same way as I would think that the installed library and header files are the primary artifact of this repo for developers.)

lonestar/CMakeLists.txt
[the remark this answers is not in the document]
roshandathathri: It's easier if I list what I would prefer as suffix for each directory:

- lonestar/analytics/cpu : *-cpu
- lonestar/analytics/gpu : *-cpu
- lonestar/analytics/distributed : *-dist
- lonestar/eda/cpu : *-cpu 
- lonestar/mining/cpu : *-cpu (use *-mining-cpu only if there are conflicts)
- lonestar/mining/gpu : *-gpu (use *-mining-gpu only if there are conflicts)
- lonestar/scientific/cpu : *-cpu
- lonestar/scientifc/gpu : *-gpu

lonestar/CMakeLists.txt
nicelhc13: OK

lonestar/analytics/CMakeLists.txt
nicelhc13: Agree

lonestar/CMakeLists.txt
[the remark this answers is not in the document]
insertinterestingnamehere: > * lonestar/analytics/gpu : *-cpu
I think you meant `*-gpu` here. Other than that, yes, this is right.

And, just to clarify, the suffixes are not for the app directory names. That info is already present in the directory names and we don't need to add it twice.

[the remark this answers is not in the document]
roshandathathri: Yes! Apps in each directory, not the directory itself. Thanks.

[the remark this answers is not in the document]
nicelhc13: Cool. Thanks

nicelhc13:
I separated CMake codes of all the CPU apps, but there is still CMake functions on the GPU and the Dist apps. I considered that those would be a separate issue.

insertinterestingnamehere:
Current version looks right to me and the one test failure is unrelated. I'll merge this soon unless anyone else has anything else to say. Most issues can be handled elsewhere to avoid blocking everything else any longer.

PR #285, opened 2020-06-08
run test for bc-level and bc-async

l-hoang:
Adds a test for bc-level and bc-async.

Note this will likely conflict with the huge renaming change, so I will not merge until after that is done and I reconcile any naming differences.

### DIFF
--- a/lonestar/analytics/cpu/betweennesscentrality/BetweennessCentralityOuter.cpp
+++ b/lonestar/analytics/cpu/betweennesscentrality/BetweennessCentralityOuter.cpp
@@ -36,6 +36,11 @@ static const char* url = "betweenness_centrality";
 static llvm::cl::opt<std::string> inputFile(llvm::cl::Positional,
                                             llvm::cl::desc("<input file>"),
                                             llvm::cl::Required);
+static llvm::cl::opt<unsigned int> numOfSources(
+    "numOfSources",
+    llvm::cl::desc("Number of sources to compute BC on (default "
+                   "is all); takes precedence over numOfOutSources"),
+    llvm::cl::init(0));
 static llvm::cl::opt<unsigned int>
     iterLimit("numOfOutSources",
               llvm::cl::desc("Number of sources WITH EDGES "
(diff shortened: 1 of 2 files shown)


roshandathathri:
Why not use rmat15 with 4 sources so that it is consistent with distributed BC tests?

l-hoang:
No particular reason. I'll switch it over.

l-hoang:
Changed it.

BC-Outer is may not be doing the same thing as the others test wise, however. It will always only iterate over vertices with edges and ignores vertices without them.

In rmat15's case the first four vertices are the ones with the highest outdegree, so it should end up doing the same calculation for this particular test.

roshandathathri:
Thanks, Loc. LGTM

l-hoang:
Made one more change to CL option used by bc-outer: "limit" and its description were very unintuitive.

lonestar/analytics/cpu/betweennesscentrality/BetweennessCentralityOuter.cpp
```diff
@@ -36,10 +36,11 @@ static const char* url = "betweenness_centrality";
 static llvm::cl::opt<std::string> inputFile(llvm::cl::Positional,
                                             llvm::cl::desc("<input file>"),
                                             llvm::cl::Required);
-static llvm::cl::opt<int> iterLimit("limit",
-                                    llvm::cl::desc("Limit number of iterations "
-                                                   "to value (0 is all nodes)"),
-                                    llvm::cl::init(0));
+static llvm::cl::opt<unsigned int>
+    iterLimit("numOfOutSources",
+              llvm::cl::desc("Number of sources WITH EDGES "
```
roshandathathri, line 41: I would numOfSources to be consistent with other apps. I would prefer with/without edges to be consistent across apps.

l-hoang: Not all BC apps support numOfOutSources at the moment. I can see about adding that support in to the ones that don't have it.

Outer doesn't have numOfSources as it doesn't make sense to give a thread empty work (outer assigns individual sources to threads).

gurbinder533: Can we also add a command-line argument to pass multiple sources instead of just reading from the file? For example, -numOfOutSources=4 -sources=2,3,44,5.

l-hoang: opened at #286

roshandathathri: > I would prefer numOfSources to be consistent with other apps. I would prefer with/without edges to be consistent across apps.

That's true for other apps as well. Just be consistent among all of them. For now, if I'm fine if it considers vertices without edges.

PR #290, opened 2020-06-09
Add sssp and cc in Cython

gurbinder533:


libgalois/include/galois/Constants.h
```diff
@@ -0,0 +1,40 @@
+#include <iostream>
```
ddn0, line 1: Is it possible to put this code somewhere more specific to the Python bindings?

I don't mind using adding Python bindings as an opportunity to clean up the C++ API but adding `UpdateRequestIndexer` seems too application-specific add to the general C++ library.

python/galois/CMakeLists.txt
```diff
@@ -16,8 +16,13 @@ add_library(_pagerank MODULE ${PAGERANK_SOURCES})
 python_extension_module(_pagerank)
 target_link_libraries(_pagerank Galois::shmem)
 
+add_cython_target(_connectedComponents _connectedComponents.pyx shmem.pyx CXX OUTPUT_VAR CONNECTEDCOMPONENTS_SOURCES)
```
ddn0, line 19: I think once we enter python land, components should have snake_case names (if we have a choice) to match the convention in that ecosystem.

python/galois/_connectedComponents.pyx
```diff
@@ -0,0 +1,105 @@
+# cython: cdivision= True
```
ddn0, line 1: I didn't look to closely to the python code since I'm imagining this more as PoC code rather than anything else. Let me know if you want me to pay closer attention.

python/galois/CMakeLists.txt
insertinterestingnamehere: This.

insertinterestingnamehere: Just to elaborate on this point, in Python, an underscore prefix is commonly used to denote internal APIs that library developers don't intend end-users to ever see or use.

libgalois/include/galois/Constants.h
gurbinder533: I agree that UpdateRequestIndexer is too application-specific but this is the best I could do to make Cython generate correct code for OBIM. I will be working with Arthur to find better way to do this.

python/galois/_connectedComponents.pyx
gurbinder533: This is still a PoC for now. I just wanted to get all this code in for Arthur to start looking at to get started.

libgalois/include/galois/Constants.h
ddn0: I'd personally lean towards keeping this on a branch (even in this repo) until we can resolve this. We are probably not that far away from doing this in a better way, any collaboration only involves you and Arthur, and `Constants.h` doesn't seem like a good (even if temporary place) to put this code in `libgalois`.

gurbinder533: I created libpygalois for cython/python related stuff. Is that fine? If not then we can move this to a separate branch.

ddn0: `libpygalois` seems fine to me; I know that @arthurp was looking into some initialization issues that might require rearranging the Cython build, what ever he comes up with might also be a good place to put shared code between Python modules. But you can probably figure this out later.

PR #293, opened 2020-06-10
Fix memory leak in all dist-apps

bozhiyou:
The issue seems like the missing `delete` of the DistGraph pointer, which is initialized by distGraphInitialization where a `new` nests, and thus the MRBCTree in NodeData is not finally destructed. A further issue is that `galois::graphs::DistGraph`, as an abstract base class, does not have a virtual destructor; when a point of DistGraph is instantiated as a derived `NewDistGraphGeneric`, [the behavior on `delete` is undefined](https://stackoverflow.com/questions/39632563/what-happens-when-delete-a-polymorphic-object-without-a-virtual-destructor).

However, bc_level surprisingly passed all the tests although there's also no `delete` with it, so I'm not sure if I found the essential cause of it. If it is, a better fix should be replacing the raw pointer of DistGraph with smart pointers like `std::unique_ptr`. I assume other dist-apps shall have the same issue but may not fail the test since the result is undefined.


roshandathathri:
After discussion (offline), let's go ahead with the virtual destructor (for now).

Bozhi, please use unique_ptr instead of explicit delete in all dist apps and then we can merge this.

libcusp/include/galois/graphs/CuSPPartitioner.h
```diff
@@ -38,6 +38,10 @@ enum CUSP_GRAPH_TYPE {
   CUSP_CSC  //!< Compressed sparse column graph format, i.e. incoming edges
 };
 
+template <typename NodeData, typename EdgeData>
+using DistGraphPtr =
```
l-hoang, line 42: See comment in Input.h

lonestar/libdistbench/include/DistBench/Input.h
```diff
@@ -116,6 +116,10 @@ extern cll::opt<std::string> mastersFile;
  * Graph-loading functions
  ******************************************************************************/
 
+template <typename NodeData, typename EdgeData>
+using DistGraphPtr =
```
l-hoang, line 120: This using typedef is defined in CuspPartitioner.h which this file includes. The conflict of names doesn't seem to be causing issues, but it's something to keep in mind. You might be able to just reuse the one defined in CusPPartitioner.h to avoid redeclaration.

lonestar/libdistbench/include/DistBench/Start.h
```diff
@@ -117,6 +117,15 @@ static void marshalGPUGraph(
 }
 #endif
 
+template <typename NodeData, typename EdgeData>
+using DistGraphPtr =
```
l-hoang, line 121: Same comment in Input.h; this includes input.h which includes CuSPPartitioner.h

lonestar/libdistbench/include/DistBench/Start.h
```diff
@@ -117,6 +117,15 @@ static void marshalGPUGraph(
...
+    std::unique_ptr<galois::graphs::DistGraph<NodeData, EdgeData>>;
+template <typename Graph>
+using SubstratePtr = std::unique_ptr<galois::graphs::GluonSubstrate<Graph>>;
```
l-hoang, line 124: This seems to be unused; if so, remove it.

Note that not any graph will work with GluonSubstrate (there's a set of functions GluonSubstrate will look for), so this probably won't even work depending on what Graph is defined as.

lonestar/libdistbench/include/DistBench/Input.h
bozhiyou: They are just for brevity. I'll use `auto` to replace the outer layers and only keep it in where the core function resides.

lonestar/libdistbench/include/DistBench/Start.h
bozhiyou: A debug relic lol but thanks for pointing this out :p

lonestar/libdistbench/include/DistBench/Input.h
l-hoang: I wouldn't use auto if you can just be explicit about it; makes the code way easier to understand.

What I'm saying is that you don't need to declare DistGraphPointer again because it's already declared in CuSPPartitioner; you can just reuse it if I understand correctly.

bozhiyou: That's true to avoid `auto` abuse; I just wouldn't like to create dependencies on included files by the typedef, since redeclaration does not irritate the compiler.

I prefer to keep this redundancy then, which can definitely be removed later.

l-hoang: Works for me. @roshandathathri any comments?

roshandathathri: LGTM

libcusp/include/galois/graphs/MiningPartitioner.h
```diff
@@ -50,7 +50,7 @@ class MiningGraph : public DistGraph<NodeTy, EdgeTy> {
   std::unique_ptr<Partitioner> graphPartitioner;
 
   uint32_t G2LEdgeCut(uint64_t gid, uint32_t globalOffset) const {
-    assert(isLocal(gid));
+    assert(this->isLocal(gid));
```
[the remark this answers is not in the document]
bozhiyou: Is it more efficient?

libcusp/include/galois/graphs/DistributedGraph.h
```diff
@@ -530,25 +530,38 @@ class DistGraph {
 
   std::vector<std::vector<size_t>>& getMirrorNodes() { return mirrorNodes; }
 
+private:
+  virtual unsigned get_host_id_impl(uint64_t) const = 0;
```
l-hoang, line 534: Nitpick: I prefer camel case, especially since a majority of the names in this file are camel case.
Another thing I had been wanting to do for years is to make everything in libdist and friends camel case, but that would require quite a bit of refactoring.

Keeping it consistent where we can should be prioritized.

libcusp/include/galois/graphs/MiningPartitioner.h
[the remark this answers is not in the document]
l-hoang: Not necessarily, but it's way more clear to a dev reading the code where the function is coming from.

libcusp/include/galois/graphs/MiningPartitioner.h
```diff
@@ -1154,7 +1151,8 @@ class MiningGraph : public DistGraph<NodeTy, EdgeTy> {
     }
   }
 
-  virtual bool is_vertex_cut() const { return false; }
+private:
```
l-hoang, line 1154: Move this to be close to the other impl functions in the code so a dev doesn't have to jump around the file looking for them. It also makes it so you don't have to declare private again.

libcusp/include/galois/graphs/MiningPartitioner.h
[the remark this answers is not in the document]
bozhiyou: Make sense. Will change.

libcusp/include/galois/graphs/DistributedGraph.h
bozhiyou: I do agree with the style consistency! However, it's difficult to refactor as you said. I personally follow [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Variable_Names), but will change it to camel case as the original file adopts.

libcusp/include/galois/graphs/MiningPartitioner.h
bozhiyou: I intentionally kept it where it was to make the changes clear for your review :p

Now that the change is acceptable, I'll make it compact.

libcusp/include/galois/graphs/DistributedGraph.h
l-hoang: Yes, at the moment the Galois codebase is pretty inconsistent with the function casing.

l-hoang: As a side note, I vastly prefer camel casing (because underscores are very un-ergonomic: underscore key is typically located in a place that needs the pinky to hit, and making the pinky do more work is terrible on your hands since it's the weakest finger on your hand).

Style wise the underscores make vars easier to read, but the ergonomic tradeoff is bleh.

libcusp/include/galois/graphs/MiningPartitioner.h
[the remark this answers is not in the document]
roshandathathri: It's not about performance. It's a stylistic choice. "this->" is redundant. The only reason you need it in this case is because the function is a member of the base class, so it is better to make that clear. Moreover, it is consistent with the rest of this class where "baseClass:": is used instead of "this->"

libcusp/include/galois/graphs/DistributedGraph.h
roshandathathri: I agree with Loc but ignore this change for now. It is fine as is. It's not a priority right now.

libcusp/include/galois/graphs/MiningPartitioner.h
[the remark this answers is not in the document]
bozhiyou: Not really; without `this->`, `assert` is surprisingly not able to recognize the identifier `isLocal`. I'll change it to the base class along with other places used to use `this->` in the file.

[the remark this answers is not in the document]
insertinterestingnamehere: Idk. At least for me, the `this` is often helpful because it clarifies that a method is being called and not just some function somewhere. Whether or not to include the `this` pointer explicitly and whether or not to include the base class name doesn't strike me as something worth standardizing though.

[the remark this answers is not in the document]
l-hoang: Agree with Ian; if the base class isn't used to prefix it like we're suggesting, then this-> should be used for the reason Ian wrote above (otherwise it's unknown where the function comes from, esp. if there are multiple headers).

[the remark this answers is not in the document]
bozhiyou: I used to think `this->` is the default but clang asked me to explicitly tell where the function comes from. For the sake of dev I'll prefix it with the base class.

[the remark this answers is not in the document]
roshandathathri: Either "baseClass::" or "this->" is required for correctness. @bozhiyou are you saying former is giving compilation errors?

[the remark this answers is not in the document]
bozhiyou: Either is fine. No prefix caused errors as the code used to be.

libcusp/include/galois/graphs/MiningPartitioner.h
bozhiyou: `is_vertex_cut` is originally using underscores so I would not change it to the camel case for now.

l-hoang: That's fine

bozhiyou:
Thanks for all your reviews. The last commit (hopefully) is built and tested well for both CPU and GPU configurations locally. As a reminder, shared memory apps may also need the pointer evolution.

libcusp/include/galois/graphs/DistributedGraph.h
```diff
@@ -68,6 +68,29 @@ class DistGraph {
 
   using GraphTy = galois::graphs::LC_CSR_Graph<NodeTy, EdgeTy, true>;
 
+  // vector for determining range objects for master nodes + nodes
+  // with edges (which includes masters)
+  //! represents split of all nodes among threads to balance edges
+  std::vector<uint32_t> allNodesRanges;
```
l-hoang, line 74: I'm assuming that child classes don't actually directly access these. If so, moving them to private is a good idea. Thanks.

l-hoang:
Should be fine to merge after CI checks through it.

PR #295, opened 2020-06-10
updated README (eda)

uditagar:
I updated the README files in the eda directory


uditagar:
I updated it to match the format

PR #300, opened 2020-06-10
adding output code for bipart

uditagar:
I have added output functionality for bipart.


insertinterestingnamehere:
I think all the test failures were happening because the testing machines were failing to pull in a submodule. My best guess is that this was caused by a recent update to a submodule version that may not be reflected here yet. I've merged master into this branch to hopefully get the tests to run.

PR #305, opened 2020-06-11
modergpu switched to master branch to for cuda capability >= 7.0

vishweshjatala:
ModernGPU branch is updated to use master branch.


vishweshjatala:
Yes. I tested locally upto 6 GPUs.

Issue #204, opened 2020-05-08
Prefix all galois-specific CMake options with GALOIS


ddn0:
Question: should we bother supporting `USE_EXP` or `USE_STRICT_CONFIG` or `USE_EMBEDDED_REVISION` at all?

Also: remember to update all the scripts, ci, etc.

roshandathathri:
USE_EXP may be okay - not sure if we are using it (after Ian's changes to experimental).

I haven't used the other two. Better to drop them unless there is a specific need.

PR #324, opened 2020-06-15
Prefix CMake Argument names with GALOIS

insertinterestingnamehere:
Fixes https://github.com/IntelligentSoftwareSystems/Galois/issues/204. After a quick review of the remaining issues I figured it'd be nice to get this one done for the release since it was an easy fix.


roshandathathri:
Thanks, Ian. One minor nit - consistency between test threads and test GPUs. Either user "num" for both or drop for both?

insertinterestingnamehere:
I kept it in the GPU case because it seemed slightly ambiguous and could be interpreted like the `CUDA_VISIBLE_DEVICES`, but we're not able to handle comma separated lists of device ids. For the CPU, threads are already not tied to specific cores, so we're clearly not talking about some sort of affinity and there isn't really ambiguity.

insertinterestingnamehere:
I guess `GALOIS_NUM_TEST_THREADS` is fine though. Not a huge deal either way.

insertinterestingnamehere:
Updated. It's `GALOIS_NUM_TEST_THREADS` now.

roshandathathri:
Thanks, Ian!

PR #325, opened 2020-06-18
Add support library

ddn0:
Rebase of https://github.com/KatanaGraph/katana/pull/75 for upstream :heart: 

Biggest design question was the addition of custom formatting for error codes in Logging.h:
https://github.com/IntelligentSoftwareSystems/Galois/pull/325/files#diff-c6650dc73c3e5f6efb29ab8aaf58bdf1R20

It might be a little unexpected that the linking of std::error_code and fmt happens in `Logging.h` but I couldn't think of a better place.

### DIFF
--- a/libsupport/CMakeLists.txt
+++ b/libsupport/CMakeLists.txt
@@ -1,0 +1,43 @@
+add_library(galois_support STATIC)
+add_library(Galois::support ALIAS galois_support)
+set_target_properties(galois_support PROPERTIES EXPORT_NAME support)
+add_dependencies(lib galois_support)
+
+set(sources
+        src/GetEnv.cpp
+        src/Logging.cpp
+)
+
+target_sources(galois_support PRIVATE ${sources})
+
+target_include_directories(galois_support PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+  $<INSTALL_INTERFACE:include>
+)
+
+find_package(fmt REQUIRED)
+if (fmt_VERSION VERSION_LESS 4)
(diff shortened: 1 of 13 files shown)


libsupport/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+add_dependencies(lib galois_support)
+
+set(sources
+        src/GetEnv.cpp
+        src/Logging.cpp
+)
+
+target_sources(galois_support PRIVATE ${sources})
+
+target_include_directories(galois_support PUBLIC
```
insertinterestingnamehere, line 13: Given that we only support fmt 6.0 and later, we should check for that.

libsupport/src/Logging.cpp
```diff
@@ -0,0 +1,43 @@
...
+  }
+  o << s << "\n";
+  if (flush) {
+    o.flush();
+  }
+}
+
+} // end unnamed namespace
+
+void galois::internal::LogString(galois::LogLevel level, const std::string& s) {
```
insertinterestingnamehere, line 27: +1 to configuring the output stream here. That'll make OS compatibility stuff easier to handle later.

libsupport/CMakeLists.txt
ddn0: I actually back ported support to fmt 4.0 to support the standard fmt package in bionic. Let me update the docs and the CMake `find_packages`.

insertinterestingnamehere: Sounds good. Thanks!

ddn0:
@insertinterestingnamehere ping (this PR is waiting on your merge and updated `lmod` module).

insertinterestingnamehere:
Just built the module and confirmed that this works with our internal packages. Merging. Thanks!

PR #327, opened 2020-06-19
Change pagerank updation rule to match GAP

gurbinder533:
GAP (http://gap.cs.berkeley.edu/) is considered standard for analytics.

### DIFF
--- a/lonestar/analytics/cpu/pagerank/PageRank-pull.cpp
+++ b/lonestar/analytics/cpu/pagerank/PageRank-pull.cpp
@@ -59,11 +59,12 @@ using ResidualArray = galois::LargeArray<PRTy>;
 
 //! Initialize nodes for the topological algorithm.
 void initNodeDataTopological(Graph& g) {
+  PRTy init_value = 1.0f / g.size();
   galois::do_all(
       galois::iterate(g),
       [&](const GNode& n) {
         auto& sdata = g.getData(n, galois::MethodFlag::UNPROTECTED);
-        sdata.value = INIT_RESIDUAL;
+        sdata.value = init_value;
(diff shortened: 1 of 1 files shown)


lonestar/analytics/cpu/pagerank/PageRank-pull.cpp
```diff
@@ -217,32 +218,31 @@ void computePRTopological(Graph& graph) {
...
-    if (delta <= tolerance || iteration >= maxIterations) {
+    if (accum.reduce() <= tolerance || iteration >= maxIterations) {
       break;
     }
-    max_delta.reset();
+    accum.reset();
 
   } ///< End while(true).
 
+  std::cout << "Rounds executed: " << iteration << "\n";
```
roshandathathri, line 245: We print this line in stats right? If we don't, then include it in stats. Please remove this line.

lonestar/analytics/cpu/pagerank/PageRank-pull.cpp
```diff
@@ -63,7 +63,7 @@ void initNodeDataTopological(Graph& g) {
       galois::iterate(g),
       [&](const GNode& n) {
         auto& sdata = g.getData(n, galois::MethodFlag::UNPROTECTED);
-        sdata.value = INIT_RESIDUAL;
+        sdata.value = 1.0f / g.size();
```
roshandathathri, line 66: Declare a constant outside the do_all()

l-hoang:
question: planning to do this for distributed as well?

roshandathathri:
> question: planning to do this for distributed as well?

Our cpu, gpu, and distributed codes should be consistent with each other.

PR #329, opened 2020-06-20
Fix DistGraph::getData API to support void NodeTy reference

bozhiyou:
I'm not sure if this is on purpose to enforce a non-void NodeTy, but IMO it's better to keep consistent with the intrinsic LC_CSR_Graph APIs.

IIRC I once asked Loc why we need [a dummy NodeData in dist-tc](https://github.com/IntelligentSoftwareSystems/Galois/blob/master/lonestar/analytics/distributed/triangle-counting/tc.cpp#L56) but forgot the answer lol. I think in topology mining apps it's totally avoidable, and maybe we should derive a graph type with no labels in nodes and edges (not LC but just CSR). The current [PangolinGraph ](https://github.com/IntelligentSoftwareSystems/Galois/blob/master/libpangolin/include/pangolin/gtypes.h#L34) also enforces `uint32_t` NodeTy.

BTW I wonder why the dist-tc app resides in analytics/distributed but not mining/distributed?


roshandathathri:
Triangle counting is there is lonestar analytics for cpu, gpu, and distributed. The algorithm using for triangle counting in mining is different and other mining algorithms are more complicated than triangle counting.

bozhiyou:
> Triangle counting is there is lonestar analytics for cpu, gpu, and distributed. The algorithm using for triangle counting in mining is different and other mining algorithms are more complicated than triangle counting.

To extend mining/cpu apps like kcl to distributed, should I put it in analytics/distributed or mining/distributed (not created yet)?

bozhiyou:
> If this builds/runs fine, it's fine to merge; I just have a strange nagging feeling that this change may have caused some issue at some point in the past in some use case.

Before merging this simple change, let me try to modify the current dist-tc with void NodeTy and see if the tests pass; if so, I'll commit that change here as well. Is that good to you, @l-hoang ?

l-hoang:
Sure; in fact that's probably the best test case.

roshandathathri:
> To extend mining/cpu apps like kcl to distributed, should I put it in analytics/distributed or mining/distributed (not created yet)?

mining/distributed is the right place for it. You should be using the same algorithm (on each CPU) as that in mining/cpu

l-hoang:
Do tests pass + is correctness fine (i.e. correct number of triangles returned?)? If so, please merge.

bozhiyou:
Tests passed and tried one of the test commands manually and it's correct (doesn't the test contain sanity check?).

Only cpu and distributed versions under analytics are refined; not familiar with GPU code and didn't find the corresponding part of code. I planned to modify mining engine as well, where the NodeTy is fixed to be `uint32_t`, but that could be another pr.

roshandathathri:
> doesn't the test contain sanity check?

The CI tests don't verify the output yet. Please run "make test" on a local UT machine for distributed apps (you don't have to run on GPUs if it doesn't affect GPU).

l-hoang:
TC doesn't have a correctness check: you have to check the number of triangles yourself.

bozhiyou:
I assume `make test` and `ctest` are identical. Only tried one of the tests (command line obtained by `ctest -V`) manually and it's correct. I'll try more; if any issue I'll fix it or just rollback. Thanks for the review!

roshandathathri:
> I assume `make test` and `ctest` are identical

Yes, they are

PR #334, opened 2020-06-25
Consistent graph reading bc, degree interface for file graph

l-hoang:
- Made bc async use filegraph for reading like the other 2 variants
- added a degree interface for file graph for use in graph analysis before execution

In retrospect this should have been 2 different PRs.

### DIFF
--- a/libgalois/include/galois/graphs/FileGraph.h
+++ b/libgalois/include/galois/graphs/FileGraph.h
@@ -98,6 +98,10 @@ private:
   //! adjustments to edge index when we load only part of a graph
   uint64_t edgeOffset;
 
+  //! If initialized, this array stores node degrees in memory for fast access
+  //! via the getDegree function
+  LargeArray<uint64_t> node_degrees;
+
   // graph reading speed variables
   galois::GAccumulator<uint64_t> numBytesReadIndex, numBytesReadEdgeDst,
       numBytesReadEdgeData;
@@ -440,6 +444,20 @@ public:
     EdgeTy* r = reinterpret_cast<EdgeTy*>(edgeData);
     return &r[numEdges];
   }
+
+  //! Calculates node degrees and saves them to a class variable for
+  //! access by getDegree.
+  void initNodeDegrees();
+
+  /**
+   * Gets the degree of a particular node. Assumes that initNodeDegrees has
+   * been called to initialize the array of degrees; if not, it is very likely
+   * going to segfault as it will attempt to access uninitialized memory.
+   *
+   * @param node_id ID of a node to get the degree of.
+   * @returns Degree of a specified node
+   */
+  uint64_t getDegree(uint32_t node_id) const;
 
   /**
    * Gets the first node of the loaded graph.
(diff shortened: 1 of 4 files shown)


libgalois/src/FileGraph.cpp
```diff
@@ -708,5 +708,39 @@ FileGraph::iterator FileGraph::end() const {
...
+          } else {
+            this->node_degrees.set(n, this->outIdx[0]);
+          }
+        },
+        galois::loopname("FileGraphInitNodeDegrees"), galois::no_stats());
+  }
+}
+
+uint64_t FileGraph::getDegree(uint32_t node_id) const {
+  if (this->node_degrees.size()) {
```
roshandathathri, line 731: Add an assertion for this and change the documentation. This can be called within a parallel function. We should not be calculating it on-the-fly.

libgalois/include/galois/graphs/FileGraph.h
```diff
@@ -98,6 +98,10 @@ class FileGraph {
   //! adjustments to edge index when we load only part of a graph
   uint64_t edgeOffset;
 
+  //! If initialized, this array stores node degrees in memory for fast access
+  //! via a getDegree function
+  LargeArray<uint64_t> node_degrees;
```
roshandathathri, line 103: Instead of node degrees, I would prefer if it is the prefix sum array (indices) of the CSR directly so that it can be used for loading edges subsequently.

lonestar/analytics/cpu/betweennesscentrality/BetweennessCentrality.cpp
```diff
@@ -100,6 +100,11 @@ int main(int argc, char** argv) {
   galois::StatTimer totalTime("TimerTotal");
   totalTime.start();
 
+  galois::graphs::FileGraph degreeGraph;
+  degreeGraph.fromFile(inputFile);
```
roshandathathri, line 104: Can you pass this loaded graph to the 3 variants so that they can load the edges faster?

libgalois/src/FileGraph.cpp
l-hoang: in other words, make it getDegree fails if init isn't called beforehand?

lonestar/analytics/cpu/betweennesscentrality/BetweennessCentrality.cpp
l-hoang: for 2 of the 3, no, because those use the readGraph interface (which I am not familiar with)

for 1 of the 3, yes, but that would require changing FileGraph to detect if the prefix sum (assuming that's what I change the node_degrees above to, which is debatable given what I'm about to write here) is loaded: if it is, it uses the prefix sum array instead of accessing the mmap'd prefix sum on disk. this would require quite a bit of code change

roshandathathri: Ok, makes sense. This is fine for now.

libgalois/src/FileGraph.cpp
roshandathathri: Yes.

l-hoang:
Added an assertion and fixed the documentation.

l-hoang:
Travis build failure is actually a failure this time around; will fix.

Issue #272, opened 2020-06-04
Use GALOIS_ASSERT and gPrint in libpangolin and mining instead of assert() and printf


roshandathathri:
See #268 for some discussion.

l-hoang:
This would require some major refactoring of the code I think, but yes, it should be done.

PR #337, opened 2020-06-27
replace std::cout with gPrint

chenxuhao:
fix issue #272
https://github.com/IntelligentSoftwareSystems/Galois/issues/272


roshandathathri:
@chenxuhao The code looks good. I have 2 questions:

(1) SGL readme and help seems to indicate that any pattern can be queried but the code seems to be specific for cycle and diamond patterns. Can you please clarify?

(2) Regarding #272, I did not see any assertion when k is invalid (check #268)?

lonestar/mining/cpu/subgraph-listing/README.mb
```diff
@@ -0,0 +1,37 @@
...
+This application counts the occurances of a given subgraph in a graph. 
+Currently only two patterns are supported: diamond and 4-cycle.
+
+INPUT
+--------------------------------------------------------------------------------
+
+This application takes in symmetric and simple Galois .gr graphs.
+You must specify both the -symmetricGraph and the -simpleGraph flags when
+running this benchmark.
+You must also specify the query graph (i.e. pattern) using -p.
```
roshandathathri, line 16: Please remove this line.

chenxuhao: this is required

roshandathathri: Isn't the query graph a cycle or a diamond depending on the executable? I don't understand what needs to be specified. Can you please add an example?

chenxuhao: Examples are already there. The commandline format is designed for any arbitrary pattern. The current implementation is ad-hoc so that the executables are limited to a specific pattern. All executables will be merged into one soon, in which case -p is used to specify the pattern.

roshandathathri: > The current implementation is ad-hoc so that the executables are limited to a specific pattern

That is my point. When you merge into one executable, you can reintroduce that line in the README. For now, the README and the executables are not consistent with each other.

chenxuhao: which part is not consistent?

Issue #78, opened 2020-03-23
Add Adaptive OBIM Worklist


insertinterestingnamehere:
In a recent SC paper (https://www.cs.tau.ac.il/~mad/publications/sc2019-cps.pdf) @serifyesil put together an adaptive version of OBIM available at https://github.com/serifyesil/pmod. It would be great if we could get a version of this new worklist up and running against the newer version of Galois.

insertinterestingnamehere:
Note: Since Galois has changed license, we need explicit permission to release the updated code under the BSD license instead of the old UT Research License. @serifyesil, is the new licensing okay with you?

insertinterestingnamehere:
@serifyesil? Can we adapt your code for the newer Galois release and release it under the BSD license instead of the old license?

serifyesil:
BSD license is okay for me. Feel free to use the code.

insertinterestingnamehere:
Thank you!

insertinterestingnamehere:
I've been working on getting this code working with the current version of Galois, but given that it's a new feature, I'm deferring that until after the release is finalized.

PR #342, opened 2020-07-02
Port Adaptive OBIM (PMOD) to Latest Galois

insertinterestingnamehere:
Fixes https://github.com/IntelligentSoftwareSystems/Galois/issues/78.

Very much still a work in progress. A relatively major overhaul is needed to get this up-to-date since the original PMOD scheduler was written against Galois 2.2.1 and so much has changed since then.

### DIFF
--- a/libgalois/include/galois/worklists/AdaptiveObim.h
+++ b/libgalois/include/galois/worklists/AdaptiveObim.h
@@ -537,7 +537,7 @@ public:
 
     CTy* C = p.current;
     if (BlockPeriod &&
-        (BlockPeriod < 0 || (p.numPops++ & ((1ull << BlockPeriod) - 1) == 0)))
+        (BlockPeriod < 0 || ((p.numPops++ & ((1ull << BlockPeriod) - 1)) == 0)))
       return slowPop(p);
 
     galois::optional<value_type> retval;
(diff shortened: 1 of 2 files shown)


insertinterestingnamehere:
Okay, I think the header for the new work list is ready to go. We need code that exercises it, so I'm planning on adding their implementation of A* search as well, but that can be done in a separate PR. Currently we're just testing that it compiles okay. Once we finalize this I'm planning on squashing all the commits since they're currently a bit of a mess.

insertinterestingnamehere:
@bozhiyou FYI after looking through this code it appears that it actually still only handles integer priorities. For now to get floating point priorities we'll still have to do some sort of really conservative round-off scheme. I think their adaptive binning approach still can work with floats, but that'll take some more work beyond what's already here.

insertinterestingnamehere:
With the exception of the changed parentheses in the last commit, this PR is not intended to significantly change the gehavior of the original adaptive OBIM scheduler. It's mainly focused at porting the original PMOD/Adaptive OBIM scheduler to the newer version of Galois. For reference, the original is available at  https://github.com/serifyesil/PMOD/blob/master/Galois-2.2.1/include/Galois/WorkList/AdaptiveObim.h.

bozhiyou:
Thanks, Ian! I'm finalizing the code on my side and will check this out soon.

insertinterestingnamehere:
@bozhiyou This is actually usable in apps now, though it'd definitely still benefit from a lot more testing. Please try it out and let me know if you run into any more bugs.

PR #345, opened 2020-07-07
Lower-level FGW w/ careful memory management

bozhiyou:
The current FileGraphWriter does wild memory management: to construct a new graph (from an old graph, e.g. make symmetric, or not), it first builds up `std::vector`s in CSR format, then `memcpy` them to `mmap`'ed space, then reads the data back to the fields inherited from FileGraph, and finally constructs the new graph by the FileGraph. Memory usage is double the size of the new graph, which is bad esp. when the graph is large (in my case, I'm synthesizing regular grids).

I initiated a new FileGraphWriter implementation, namely FileGraphWeaver so that it may co-exist with the old one or replace the old FGW later, which has the same APIs (means least efforts on other code if it replaces the old impl.) but does mmap directly without memcpy.

---

EDIT: The changes are in-place now.

As a reminder for discussion: Should FileGraphWriter be a `friend` instead of a child of FileGraph? They in fact share the same data fields which are private in FileGraph and re-declared in FileGraphWriter. Making these fields protected is a solution but does not seem to be a good practice just in order to expose them to a specific subclass.

### DIFF
--- a/libgalois/include/galois/graphs/FileGraph.h
+++ b/libgalois/include/galois/graphs/FileGraph.h
@@ -48,6 +48,15 @@
 #include "galois/substrate/NumaMem.h"
 #include "galois/Reduction.h"
 
+template <typename T>
+struct SizeOf {
+  static constexpr std::size_t value = sizeof(T);
+};
+template <> // sizeof never returns 0
+struct SizeOf<void> {
+  static constexpr std::size_t value = 0;
+};
+
 namespace galois {
 namespace graphs {
 
@@ -720,9 +729,9 @@ public:
 
   //! Adds a neighbor between src and dst w/ corresponding data
   template <typename T>
-  size_t
-  addNeighbor(size_t src, size_t dst,
-              typename std::enable_if<!std::is_void<T>::value>::type&& data) {
+  size_t addNeighbor(
+      size_t src, size_t dst,
+      const typename std::enable_if<!std::is_void<T>::value, T>::type& data) {
     assert(edgeData);
     size_t idx                          = addNeighbor(src, dst);
     reinterpret_cast<T*>(edgeData)[idx] = data;
@@ -753,11 +762,8 @@ public:
 template <typename EdgeTy>
 void makeSymmetric(FileGraph& in_graph, FileGraph& out) {
   typedef FileGraph::GraphNode GNode;
-  typedef LargeArray<EdgeTy> EdgeData;
-  typedef typename EdgeData::value_type edge_value_type;
 
   FileGraphWriter g;
-  EdgeData edgeData;
 
   size_t numEdges = 0;
 
@@ -776,7 +782,7 @@ void makeSymmetric(FileGraph& in_graph, FileGraph& out) {
 
   g.setNumNodes(in_graph.size());
   g.setNumEdges(numEdges);
-  g.setSizeofEdgeData(EdgeData::has_value ? sizeof(edge_value_type) : 0);
+  g.setSizeofEdgeData(SizeOf<EdgeTy>::value);
 
   g.phase1();
   for (FileGraph::iterator ii = in_graph.begin(), ei = in_graph.end(); ii != ei;
@@ -793,7 +799,6 @@ void makeSymmetric(FileGraph& in_graph, FileGraph& out) {
   }
 
   g.phase2();
-  edgeData.create(numEdges);
   for (FileGraph::iterator ii = in_graph.begin(), ei = in_graph.end(); ii != ei;
        ++ii) {
     GNode src = *ii;
@@ -801,24 +806,20 @@ void makeSymmetric(FileGraph& in_graph, FileGraph& out) {
                                   ej = in_graph.edge_end(src);
          jj != ej; ++jj) {
       GNode dst = in_graph.getEdgeDst(jj);
-      if (EdgeData::has_value) {
-        edge_value_type& data = in_graph.getEdgeData<edge_value_type>(jj);
-        edgeData.set(g.addNeighbor(src, dst), data);
-        if (src != dst)
-          edgeData.set(g.addNeighbor(dst, src), data);
-      } else {
+      if constexpr (std::is_void<EdgeTy>::value) {
         g.addNeighbor(src, dst);
         if (src != dst)
           g.addNeighbor(dst, src);
+      } else {
+        EdgeTy& data = in_graph.getEdgeData<EdgeTy>(jj);
+        g.addNeighbor<EdgeTy>(src, dst, data);
+        if (src != dst)
+          g.addNeighbor<EdgeTy>(dst, src, data);
       }
     }
   }
 
-  edge_value_type* rawEdgeData = g.finish<edge_value_type>();
-  if (EdgeData::has_value)
-    std::uninitialized_copy(std::make_move_iterator(edgeData.begin()),
-                            std::make_move_iterator(edgeData.end()),
-                            rawEdgeData);
+  g.finish();
 
   out = std::move(g);
 }
@@ -837,16 +838,13 @@ void makeSymmetric(FileGraph& in_graph, FileGraph& out) {
 template <typename EdgeTy, typename PTy>
 void permute(FileGraph& in_graph, const PTy& p, FileGraph& out) {
   typedef FileGraph::GraphNode GNode;
-  typedef LargeArray<EdgeTy> EdgeData;
-  typedef typename EdgeData::value_type edge_value_type;
 
   FileGraphWriter g;
-  EdgeData edgeData;
 
   size_t numEdges = in_graph.sizeEdges();
   g.setNumNodes(in_graph.size());
   g.setNumEdges(numEdges);
-  g.setSizeofEdgeData(EdgeData::has_value ? sizeof(edge_value_type) : 0);
+  g.setSizeofEdgeData(SizeOf<EdgeTy>::value);
 
   g.phase1();
   for (FileGraph::iterator ii = in_graph.begin(), ei = in_graph.end(); ii != ei;
@@ -860,7 +858,6 @@ void permute(FileGraph& in_graph, const PTy& p, FileGraph& out) {
   }
 
   g.phase2();
-  edgeData.create(numEdges);
   for (FileGraph::iterator ii = in_graph.begin(), ei = in_graph.end(); ii != ei;
        ++ii) {
     GNode src = *ii;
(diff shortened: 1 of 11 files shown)


bozhiyou:
Thanks for the review!

> (1) Have you tested all uses of FileGraphWriter (whether it is correct after your changes)? While doing so, please add ctests so that they can verified in the future.

I assume the existed ctests for `graph-convert` are sufficient since it calls all the functionalities of FileGraphWriter as a standard process.

> (2) FileGraphWriter being a derived class of FileGraph is fine. The base class is reader-only whereas the derived class is reader+writer. That's a reasonable design. Please make the appropriate variables protected in the base class.

IMO FileGraph is more of a storage class rather than a reader - it's being read later via `readGraph`. FileGraphWriter is manipulating this storage class by filling the memory info. Your interpretation also makes sense to me though and the changes reflect this.

> (3) There is lack of consistency in whether a function is defined in the header or in the cpp. Unless the definition is a couple of lines, please move it to the cpp file.

It was all defined in the header file. I put especially `phase1()` to the .cpp file because it calls `mmap_big` defined there. By the way, should `mmap_big`, which switch between `mmap` and `mmap64`, be extracted to a separate header file like the one for endian stuff so that we can do mmap freely in other places when necessary?

roshandathathri:
> It was all defined in the header file. I put especially `phase1()` to the .cpp file because it calls `mmap_big` defined there.

Please move all function definitions to the cpp file unless it is a template function.

> By the way, should `mmap_big`, which switch between `mmap` and `mmap64`, be extracted to a separate header file like the one for endian stuff so that we can do mmap freely in other places when necessary?

Good idea. Yes, please!

bozhiyou:
The CI tests failed on the uninitialized pointer in libpangolin and I'm not going to fix that in this pr.

Please feedback before merging, thanks!

libgalois/include/galois/graphs/FileGraph.h
```diff
@@ -48,6 +48,15 @@
 #include "galois/substrate/NumaMem.h"
 #include "galois/Reduction.h"
 
+template <typename T>
+struct SizeOf {
+  static constexpr std::size_t value = sizeof(T);
+};
+template <> // sizeof never returns 0
```
roshandathathri, line 55: Why do you need this?

libgalois/include/galois/graphs/FileGraph.h
```diff
@@ -700,82 +699,58 @@ class FileGraphWriter : public FileGraph {
...
-    } else {
-      // version 2
-      size_t idx = base + (starts64)[src]++;
-      assert(idx < outIdx[src]);
-      outs64[idx] = dst;
-      return idx;
-    }
+  //! Adds a neighbor between src and dst w/ corresponding data
+  template <typename T>
+  size_t addNeighbor(
```
roshandathathri, line 732: You have both a templated and non-templated function with the same name. Why does this not cause a compiler error? Why not handle it by a special case of edge data being void type?

libgalois/include/galois/graphs/FileGraph.h
```diff
@@ -810,7 +782,7 @@ void makeSymmetric(FileGraph& in_graph, FileGraph& out) {
 
   g.setNumNodes(in_graph.size());
   g.setNumEdges(numEdges);
-  g.setSizeofEdgeData(EdgeData::has_value ? sizeof(edge_value_type) : 0);
+  g.setSizeofEdgeData(SizeOf<EdgeTy>::value);
```
roshandathathri, line 785: Instead of doing size, you can check if the EdgeTy is void and set it to 0.

libgalois/include/galois/graphs/FileGraph.h
roshandathathri: A better design might be to have a separate addEdgeData function which takes an index (that is returned by addNeighbor) and data.

libgalois/include/galois/graphs/FileGraph.h
```diff
@@ -871,16 +838,13 @@ void makeSymmetric(FileGraph& in_graph, FileGraph& out) {
...
-  typedef typename EdgeData::value_type edge_value_type;
 
   FileGraphWriter g;
-  EdgeData edgeData;
 
   size_t numEdges = in_graph.sizeEdges();
   g.setNumNodes(in_graph.size());
   g.setNumEdges(numEdges);
-  g.setSizeofEdgeData(EdgeData::has_value ? sizeof(edge_value_type) : 0);
+  g.setSizeofEdgeData(SizeOf<EdgeTy>::value);
```
roshandathathri, line 847: Ditto

lonestar/analytics/cpu/preflowpush/Preflowpush.cpp
```diff
@@ -585,7 +582,7 @@ struct PreflowPush {
 
     p.setNumNodes(reader.size());
     p.setNumEdges(numEdges);
-    p.setSizeofEdgeData(sizeof(edge_value_type));
+    p.setSizeofEdgeData(SizeOf<EdgeTy>::value);
```
roshandathathri, line 585: Ditto

libgalois/include/galois/graphs/FileGraph.h
bozhiyou: As you've seen, it's for determining the size of edge data. See my reply below for more details.

libgalois/include/galois/graphs/FileGraph.h
bozhiyou: Here's what I tried:

- `std::is_void<EdgeTy>::value ? 0 : sizeof(EdgeTy)`
This is a *runtime* conditioning and won't compile - the compiler will still try to parse `sizeof(void)` even though 0 should be returned when EdgeTy is void.

- `sizeof(std::conditional<std::is_void<EdgeTy>::value, void*, EdgeTy>)`
In the current implementation, `EdgeData::value_type`, which is identical to `LargeArray<EdgeTy>::value_type` and also aliased as `edge_data_type`, maps void as `void*` and non-void as itself; however `sizeof(void*)` is not 0.

My design logic:

1. `LargeArray<T>::has_value` maps void to false and non-void to true by template specialization `LargeArray<void>`, so template spetializaton for void should make sense.

2. `LargeArray<T>` has a nested struct `size_of`, which for void is
```
  struct size_of {
    const static size_t value = 0;
  };
```
and for non-void is
```
  struct size_of {
    const static size_t value = sizeof(T);
  };
```
This `EdgeData::size_of::value` is also used in current implementation if you went through the comparison carefully. Therefore, generalizing it outside should make sense unless there're cases where "sizeof(void)" should not be forced to 0.

3. `sizeof` is parsed at compile time and my design reflects this.

I'd like to have your comments and thoughts on the way it should be :) thanks!

libgalois/include/galois/graphs/FileGraph.h
bozhiyou: Good question :)!

Remember the reference link you showed me about the virtual stuff? I learned something about [template functions](http://www.gotw.ca/publications/mill17.htm) there as well.

If I understand the material currently and what it suggests also makes sense to you, what I learned is that *non-template overload has the first priority than the templates* - that's why this does not cause a compiler error.

The deal I'm handling: For void, addNeighbor only requires *src* and *dst* and does only adding the edge; for non-void, addNeighbor could require *src*, *dst*, and *data*, and does not only adding the edge but also adding the edge data. The non-void case should also use the addNeighbor for void case without adding the edge data when adding the edge, but it does not make sense for void case to have addNeighbor with *src*, *dst*, and *data* (therefore I did not provide this specialization).

In my design, non-template addNeighbor could be used for both void and non-void cases and it's always called prior to the template one. The template one only supports non-void cases and compiler errors for void cases.

A separate addEdgeData may not prevent users from calling it for void cases, though assertion could be made at the beginning of its definition.

roshandathathri: Thanks for the detailed explanation. This choice/design is fine, although I like the previous design of a separate addEdgeData function which takes an index (that is returned by addNeighbor) and data.

libgalois/include/galois/graphs/FileGraph.h
roshandathathri: Thanks for the detailed explanation. I'm still not convinced that defining a custom sizeof operator is the only solution available.

> This is a _runtime_ conditioning and won't compile

Yes, have you tried using std::enable_if instead? Here's one suggestion. Let setSizeofEdgeData be a template function with no arguments. If the type is void, then it's value is set to 0; otherwise, it is set of sizeof(type). This can be achieved using two enable_if functions.

libgalois/include/galois/graphs/FileGraph.h
bozhiyou: I'll add the separate addEdgeData function as an alternative but still keep my choice :) Thanks!

roshandathathri: No, don't do that! Go with one of them.

Typically, there should NOT be two ways to do the same thing.

libgalois/include/galois/graphs/FileGraph.h
bozhiyou: In principle, I preserve the APIs as much as possible when making changes to implementations without discussing the design of the APIs.

Now that a different API is considered, I agree with your scheme and have a further thought: Instead of making the old `setSizeofEdgeData(<size>)` a new `setSizeofEdgeData<EdgeTy>()` with ugly trailing parathesis, does it make sense to combine this to `setNumEdges(<numEdges>)` as `setNumEdges<EdgeTy>(<numEdges>)`, since `setNumNodes` `setNumEdges` `setSizeofEdgeData` are always called together?

libgalois/include/galois/graphs/FileGraph.h
bozhiyou: In that case, I'd like to insist on my choice unless a more convincing objection :)

roshandathathri: Yes, that's fine.

libgalois/include/galois/graphs/FileGraph.h
roshandathathri: Yes, `setNumEdges<EdgeTy>(<numEdges>)` is a better design.

libgalois/include/galois/graphs/FileGraph.h
```diff
@@ -680,7 +671,7 @@ class FileGraph {
  *
  * Writer your file in rounds:
  * <ol>
- *  <li>setNumNodes(), setNumEdges(), setSizeofEdgeData()</li>
+ *  <li>setNumNodes(), setNumEdges()</li>
```
roshandathathri, line 674: setNumEdges<EdgeTy>()

bozhiyou: What do you mean?

roshandathathri: `setNumEdges<EdgeTy>()`

bozhiyou: I see, will fix. Thanks for the review!

libgalois/include/galois/graphs/FileGraph.h
```diff
@@ -670,112 +671,86 @@ class FileGraph {
...
  *  <li>phase1(), for each node, incrementDegree(Node x)</li>
  *  <li>phase2(), add neighbors for each node, addNeighbor(Node src, Node
```
roshandathathri, line 676: Also add other `addNeighbor<>()`

PR #346, opened 2020-07-13
Copyright Fixes

insertinterestingnamehere:
The old Texas State copyright for the GPU version of Barnes Hut was accidentally removed when we relicensed Galois. This restores the copyright. Fortunately the old license for this app was actually already BSD, so we don't have to worry about any additional relicensing issues.


insertinterestingnamehere:
@burtscher does this look good to you?

PR #348, opened 2020-07-13
Fix operator precedence bug in OBIM.

insertinterestingnamehere:
The == operator actually has higher precedence than the bitwise & operator in C++, so our handling of the BlockPeriod template parameter was actually wrong. This adds clarifying parentheses to get the right result. I'm not yet sure what the performance consequences of this change are.

I noticed this while preparing https://github.com/IntelligentSoftwareSystems/Galois/pull/342. It's a really old bug. I only found it because of a new gcc warning.

### DIFF
--- a/libgalois/include/galois/worklists/Obim.h
+++ b/libgalois/include/galois/worklists/Obim.h
@@ -405,7 +405,7 @@ public:
       return this->popStored(p, p.curIndex);
 
     if (!UseBarrier && BlockPeriod &&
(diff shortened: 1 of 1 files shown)


insertinterestingnamehere:
I ran some ad-hoc performance numbers with SSSP this afternoon. There doesn't appear to be any significant performance change with this fix, so the main benefit of this is just that the `BlockPeriod` template parameter now really does what it says it does. Merging.

Issue #349, opened 2020-07-20
Locking issue with AdaptiveObim (PMOD)


bozhiyou:
```
#0  0x00007ffff64e8387 in raise () from /lib64/libc.so.6
#1  0x00007ffff64e9a78 in abort () from /lib64/libc.so.6
#2  0x00007ffff64e11a6 in __assert_fail_base () from /lib64/libc.so.6
#3  0x00007ffff64e1252 in __assert_fail () from /lib64/libc.so.6
#4  0x0000000000411bff in galois::substrate::SimpleLock::unlock (this=0x2aaaab201080)
    at /workspace/GaloisCpp/libgalois/include/galois/substrate/SimpleLock.h:69
#5  0x0000000000411f4c in galois::substrate::PaddedLock<true>::unlock (this=0x2aaaab201080)
    at /workspace/GaloisCpp/libgalois/include/galois/substrate/PaddedLock.h:41
#6  0x0000000000446f7a in galois::worklists::AdaptiveOrderedByIntegerMetric<void FastMarching<true, galois::graphs::LC_CSR_Graph<NodeData, void, true, false, false, void>, galois::InsertBag<std::pair<double, unsigned int>, 0u>, unsigned int, std::pair<double, unsigned int> >(galois::graphs::LC_CSR_Graph<NodeData, void, true, false, false, void>&, galois::InsertBag<std::pair<double, unsigned int>, 0u>&)::{lambda(std::pair<double, unsigned int> const&)#2}, galois::worklists::internal::ChunkMaster<std::pair<double, unsigned int>, galois::worklists::ConExtLinkedStack, true, true, 32, true>, 0, true, false, 64, std::pair<double, unsigned int>, unsigned int, false, false, true>::slowPop(galois::worklists::internal::ChunkMaster<std::pair<double, unsigned int>, galois::worklists::ConExtLinkedStack, true, true, 32, true>::ThreadData&) (this=0x7fffffff8a00, p=...)
    at /workspace/GaloisCpp/libgalois/include/galois/worklists/AdaptiveObim.h:417
[...]

PR #350, opened 2020-07-23
Fix assertion failure from erroneous unlock in Adaptive OBIM.

insertinterestingnamehere:
Fixes https://github.com/IntelligentSoftwareSystems/Galois/issues/349, though I'm not entirely sure why yet.

### DIFF
--- a/libgalois/include/galois/worklists/AdaptiveObim.h
+++ b/libgalois/include/galois/worklists/AdaptiveObim.h
@@ -55,24 +55,22 @@ namespace worklists {
  *   int operator()(Item i) const { return i.index; }
  * };
  *
- * typedef galois::WorkList::AdaptiveOrderedByIntegerMetric<Indexer> WL;
+ * typedef galois::worklists::AdaptiveOrderedByIntegerMetric<Indexer> WL;
  * galois::for_each<WL>(items.begin(), items.end(), Fn);
  * \endcode
  *
- * @tparam Indexer Indexer class
- * @tparam Container Scheduler for each bucket
- * @tparam BlockPeriod Check for higher priority work every 2^BlockPeriod
- *                     iterations
- * @tparam BSP Use back-scan prevention
- * @tparam uniformBSP Use uniform back-scan prevention
- * @tparam T Work item type
- * @tparam Index Indexer return type
- * @tparam UseMonotonic Assume that an activity at priority p will not
- *                      schedule work at priority p or any priority p1
- *                      where p1 < p.
- * @tparam UseDescending Use descending order instead
- * @tparam Concurrent Whether or not to allow concurrent execution
- *
+ * @tparam Indexer        Indexer class
+ * @tparam Container      Scheduler for each bucket
+ * @tparam BlockPeriod    Check for higher priority work every 2^BlockPeriod
+ *                        iterations
+ * @tparam BSP            Use back-scan prevention
+ * @tparam uniformBSP     Use uniform back-scan prevention
+ * @tparam T              Work item type
+ * @tparam Index Indexer  return type
+ * @tparam UseMonotonic   Assume that an activity at priority p will not
+ * schedule work at priority p or any priority p1 where p1 < p.
+ * @tparam UseDescending  Use descending order instead
+ * @tparam Concurrent     Whether or not to allow concurrent execution
  */
 template <class Indexer      = DummyIndexer<int>,
           typename Container = PerSocketChunkFIFO<>, int BlockPeriod = 0,
@@ -80,27 +78,26 @@ template <class Indexer      = DummyIndexer<int>,
           typename T = int, typename Index = int, bool UseMonotonic = false,
           bool UseDescending = false, bool Concurrent = true>
 struct AdaptiveOrderedByIntegerMetric : private boost::noncopyable {
-  template <bool Concurrent_>
-  using rethread = AdaptiveOrderedByIntegerMetric<
-      Indexer, typename Container::template rethread<Concurrent_>, BlockPeriod,
-      BSP, uniformBSP, chunk_size, T, Index, UseMonotonic, UseDescending,
-      Concurrent_>;
-
-  template <typename T_>
+  template <typename _T>
   using retype = AdaptiveOrderedByIntegerMetric<
-      Indexer, typename Container::template retype<T_>, BlockPeriod, BSP,
-      uniformBSP, chunk_size, T_, typename std::result_of<Indexer(T_)>::type,
+      Indexer, typename Container::template retype<_T>, BlockPeriod, BSP,
+      uniformBSP, chunk_size, _T, typename std::result_of<Indexer(_T)>::type,
       UseMonotonic, UseDescending, Concurrent>;
 
-  template <unsigned BlockPeriod_>
+  template <bool _b>
+  using rethread = AdaptiveOrderedByIntegerMetric<
+      Indexer, typename Container::template rethread<_b>, BlockPeriod, BSP,
+      uniformBSP, chunk_size, T, Index, UseMonotonic, UseDescending, _b>;
+
+  template <unsigned _period>
   using with_block_period =
-      AdaptiveOrderedByIntegerMetric<Indexer, Container, BlockPeriod_, BSP,
+      AdaptiveOrderedByIntegerMetric<Indexer, Container, _period, BSP,
                                      uniformBSP, chunk_size, T, Index,
                                      UseMonotonic, UseDescending, Concurrent>;
 
-  template <typename Container_>
+  template <typename _container>
   using with_container =
-      AdaptiveOrderedByIntegerMetric<Indexer, Container_, BlockPeriod, BSP,
+      AdaptiveOrderedByIntegerMetric<Indexer, _container, BlockPeriod, BSP,
                                      uniformBSP, chunk_size, T, Index,
                                      UseMonotonic, UseDescending, Concurrent>;
 
@@ -228,14 +225,20 @@ private:
     unsigned int lastMasterVersion;
     unsigned int numPops;
 
-    unsigned int sinceLastFix;
-    unsigned int slowPopsLastPeriod;
-    unsigned int pushesLastPeriod;
-    unsigned int priosLastPeriod;
-
     unsigned long pmodAllDeq;
 
-    unsigned int popsFromSameQ;
+    struct {
+      unsigned int popsLastFix;
+      unsigned int slowPopsLastPeriod;
+      unsigned int pushesLastPeriod;
+      unsigned int priosLastPeriod;
+
+      unsigned int popsFromSameQ;
+
+      inline double slowPopFreq() {
+        return ((double)slowPopsLastPeriod / (double)popsLastFix);
+      }
+    } counters;
     unsigned int ctr;
 
     Index maxPrioDiffLastPeriod;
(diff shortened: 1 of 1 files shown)


bozhiyou:
The problem is actually not to do with the lock ownership. See the inline comments.

libgalois/include/galois/worklists/AdaptiveObim.h
```diff
@@ -296,8 +296,7 @@ struct AdaptiveOrderedByIntegerMetric : private boost::noncopyable {
         ((double)(p.slowPopsLastPeriod) / (double)(p.sinceLastFix)) >
             1.0 / (double)(chunk_size)) {
       for (unsigned i = 1; i < runtime::activeThreads; ++i) {
-        while (current.getRemote(i)->lock.try_lock())
```
bozhiyou, line 299: Here it is the culprit - something is missing`!`

libgalois/include/galois/worklists/AdaptiveObim.h
```diff
@@ -358,8 +357,7 @@ struct AdaptiveOrderedByIntegerMetric : private boost::noncopyable {
           delta = 0;
 
         for (unsigned i = 1; i < runtime::activeThreads; ++i) {
-          while (current.getRemote(i)->lock.try_lock())
```
bozhiyou, line 361: Ditto!

libgalois/include/galois/worklists/AdaptiveObim.h
```diff
@@ -473,8 +471,7 @@ struct AdaptiveOrderedByIntegerMetric : private boost::noncopyable {
 
   void push(const value_type& val) {
     ThreadData& p = *current.getLocal();
```
bozhiyou, line 473: The explicit lock here is fine. I prefer to start a new issue/pr for replacing some explicit lock with std::lock_guard if necessary.

libgalois/include/galois/worklists/AdaptiveObim.h
```diff
@@ -473,8 +471,7 @@ struct AdaptiveOrderedByIntegerMetric : private boost::noncopyable {
...
-    while (!p.lock.try_lock())
```
bozhiyou, line 476: The explicit lock here is fine. I prefer to start a new issue/pr for replacing some explicit lock with std::lock_guard if necessary.

libgalois/include/galois/worklists/AdaptiveObim.h
```diff
@@ -292,56 +349,50 @@ struct AdaptiveOrderedByIntegerMetric : private boost::noncopyable {
...
+          // double xx = ((double)(chunk_size) / fillRatio);
+          double xx = std::log2(chunk_size) - std::log2(numPushesThisStep) +
+                      std::log2(prioRange);
+          assert(xx);
+          delta += std::floor(xx);
+          counter <<= 1;
+        }
       }
     }
-#ifdef UNMERGE_ENABLED
```
insertinterestingnamehere, line 339: This block was originally disabled. Did you mean to re-enable it? I assume they disabled it for performance reasons.

libgalois/include/galois/worklists/AdaptiveObim.h
```diff
@@ -40,6 +39,32 @@
 namespace galois {
 namespace worklists {
 
+namespace internal {
+
+template <typename Index, bool UseDescending>
+struct AdaptiveOrderedByIntegerMetricComparator {
```
insertinterestingnamehere, line 45: When porting the code to the latest version I actually set things up to remove any reliance on the OBIM-style internal comparator classes since they really only exist to manage the direction the comparisons run. It's not a huge deal one way or the other, but I'd ultimately like to move OBIM and Adaptive OBIM away from using these helper classes since they're very verbose for what they actually are doing.

libgalois/include/galois/worklists/AdaptiveObim.h
insertinterestingnamehere: Great find. That totally explains why it was broken before. Thanks!

libgalois/include/galois/worklists/AdaptiveObim.h
```diff
@@ -228,26 +269,42 @@ struct AdaptiveOrderedByIntegerMetric : private boost::noncopyable {
...
 
+    void cleanup() {
+      popsLastFix        = 0;
+      slowPopsLastPeriod = 0;
+      pushesLastPeriod   = 0;
+
+      stats.priosLastPeriod       = 0;
+      stats.maxPrioDiffLastPeriod = 0;
+
+      minPrio = std::numeric_limits<Index>::max();
```
insertinterestingnamehere, line 294: This issue was present in the old version too, but we should use the max and min from the comparator here and not rely on the numeric_limits. A related question here though: does this scheduler even support prioritizing higher integer priorities instead of lower ones? Now that I think about it some more that may throw off some of the arithmetic elsewhere. Fortunately we're not relying on this feature right now, so it's not something we have to fix immediately.

libgalois/include/galois/worklists/AdaptiveObim.h
```diff
@@ -55,93 +80,111 @@ namespace worklists {
...
-struct AdaptiveOrderedByIntegerMetric : private boost::noncopyable {
-  template <bool Concurrent_>
+          typename T = int, typename Index = int, bool EnableUmerge = false,
+          bool UseMonotonic = false, bool UseDescending = false,
+          bool Concurrent = true>
+struct AdaptiveOrderedByIntegerMetric
+    : private boost::noncopyable,
+      public internal::AdaptiveOrderedByIntegerMetricComparator<Index,
+                                                                UseDescending> {
+  template <typename _T>
```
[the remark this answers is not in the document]
bozhiyou: That's a good point. I was just making it resemble OBIM's code without criticism on the old OBIM code. I should have discussed it with you. Thanks!

libgalois/include/galois/worklists/AdaptiveObim.h
```diff
@@ -292,56 +349,50 @@ struct AdaptiveOrderedByIntegerMetric : private boost::noncopyable {
...
+        p.isSlowPopFreq(1.0 / (double)(chunk_size))) {
+      unsigned long numPushesThisStep      = p.pushesLastPeriod;
+      unsigned long priosCreatedThisPeriod = p.stats.priosLastPeriod;
+      unsigned long allPmodDeqCounts       = p.stats.pmodAllDeq;
+      Index minOfMin                       = p.minPrio;
+      Index maxOfMax                       = p.maxPrio;
+      p.cleanup();
       for (unsigned i = 1; i < runtime::activeThreads; ++i) {
-        while (current.getRemote(i)->lock.try_lock())
+        while (!data.getRemote(i)->lock.try_lock())
```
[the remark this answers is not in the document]
bozhiyou: Actually, I'm not quite aware of the difference between try_lock and lock - Idk why they were designed separately.

libgalois/include/galois/worklists/AdaptiveObim.h
bozhiyou: I made this macro-control a template parameter, which is still false/disabled by default. To me, it doesn't make sense to keep a statically disabled code block in the codebase.

libgalois/include/galois/worklists/AdaptiveObim.h
bozhiyou: That makes sense. Again, I was matching the code structure with OBIM assuming that OBIM's code was in a good shape. Now OBIM should be the one to change.

libgalois/include/galois/worklists/AdaptiveObim.h
bozhiyou: From what I understand, this doesn't matter or have anything to do with ascending/descending - the minPrio and maxPrio are used to compute the priority range encountered in a period, which is later used to calculate the average items per priority and perform merging based on the calculation.

libgalois/include/galois/worklists/AdaptiveObim.h
```diff
@@ -357,63 +408,48 @@ struct AdaptiveOrderedByIntegerMetric : private boost::noncopyable {
...
+    // p.popsFromSameQ = 0;
 
-    updateLocal(p);
     bool localLeader = substrate::ThreadPool::isLeader();
+    deltaIndex msS(this->earliest, 0);
+
+    updateLocal(p);
 
-    deltaIndex msS;
-    msS.k = std::numeric_limits<Index>::min();
```
bozhiyou, line 391: > (quoted from comments on L294 of the last commit)
This issue was present in the old version too, but we should use the max and min from the comparator here and not rely on the numeric_limits. A related question here though: does this scheduler even support prioritizing higher integer priorities instead of lower ones? Now that I think about it some more that may throw off some of the arithmetic elsewhere. Fortunately we're not relying on this feature right now, so it's not something we have to fix immediately.

Rather, the numeric min here is problematic (but fixed now). If you look back to the initial value of min/max (earliest/identity) from the comparator, the original code wrongly made them reversed (also fixed now).

libgalois/include/galois/worklists/AdaptiveObim.h
insertinterestingnamehere: Yah, these are really easy to change. I can just take care of it in a separate PR after we get this one sorted out.

libgalois/include/galois/worklists/AdaptiveObim.h
insertinterestingnamehere: Got it. That makes sense. Thanks for the clarification.

libgalois/include/galois/worklists/AdaptiveObim.h
insertinterestingnamehere: Awesome. Thanks for catching that!

insertinterestingnamehere:
Okay, is this good to go now? It seems like the immediate outstanding issues for this PR have been taken care of.

bozhiyou:
> Okay, is this good to go now? It seems like the immediate outstanding issues for this PR have been taken care of.

I think so given that the implementation details w.r.t. OBIM will be handled in another PR. Will merge then. Please also test it on A* or whatever since I only test it with sssp to make sure nothing's error prone.

insertinterestingnamehere:
WRT subclassing, that's an interesting question. It seems like something that could maybe work, but I'm not sure whether or not we'd end up with a net simplification. For example, the UniformBSP parameter doesn't really make sense for the adaptive OBIM scheduler.

If this is in a working state, I'd like to postpone further modifications/simplifications until after the upcoming paper deadline. We've got a lot left to do for that.

insertinterestingnamehere:
Yah, I've just been testing it on SSSP and FMM too. The A* code needs to be forward ported like this was, so I'll tackle it after a few deadlines have passed.

insertinterestingnamehere:
I was just testing out the new version and now it's deadlocking in SSSP. Any idea why?

bozhiyou:
Works fine here. Could you please provide the command line?

A known issue is that gcc complains somewhere using '*' as the multiplication operator in boolean expressions which by default is dereferencing. Will commit a patch soon.

bozhiyou:
Just pushed a patch to the repo on the cdgc server but it seems not sync'ed here.

insertinterestingnamehere:
Yah, it's not automatically synchronized. If you push something to one place you should also push to the other.

insertinterestingnamehere:
https://github.com/IntelligentSoftwareSystems/Galois/pull/351 is the concrete test case I'm using.

insertinterestingnamehere:
Pushed you patch to GitHub.

bozhiyou:
Thanks! Let's move to the new PR then.

PR #352, opened 2020-08-14
Modulize mining header

bozhiyou:
For developing distributed mining apps, the bare header `lonestarmine.h` of mining apps, which contains llvm options, will be in conflict with those in libdistbench (specifically, MiningStart.h). Making the header into a module like libdistbench or liblonestar can solve this conflict.


bozhiyou:
Corresponding apps are tested locally, including gpu apps. There are some existing issues with frequent-subgraph-mining-gpu and memory-sanitizing issues with cpu mining apps, and I'll leave it there for others.

PR #354, opened 2020-08-20
Disable Louvain partitioning test in CI.

insertinterestingnamehere:


insertinterestingnamehere:
While here, I've also fixed some formatting issues that were broken by people pushing untested code.

Issue #356, opened 2020-08-21
Explicit memory order for atomic load/store


bozhiyou:
Atomic operations like `compare_exchange_weak` are mostly well called with explicit memory order, but assignments that are implicit loads/stores are not taken good care of, e.g. obtaining old values in AtomicHelpers.

Issue #246, opened 2020-05-22
Use `std::atomic` Instead of Builtin Intrinsics


insertinterestingnamehere:
We currently use the `__sync` prefixed builtins from gcc (like `__sync_fetch_and_add`) in many places. We should standardize on `std::atomic` instead. This is for both uniformity and performance. These intrinsics automatically imply sequential memory consistency and that's likely not what we need in most cases.

PR #357, opened 2020-08-21
Atomic Cleanup

insertinterestingnamehere:
The assignments and conversions with `std::atomic` have sequential consistency by default. In our use-cases, that's rarely what's actually required, so this switches all the loads, stores, and assignments to use relaxed consistency.

**If anyone is aware of anywhere where we rely on these functions using the stronger memory consistency guarantees, please say so.** A quick look at some of the uses of these helper functions didn't show anything that seemed like it would need any consistency other than relaxed. I suspect that sequential consistency was used by mistake and that it is not actually required for any of the use-cases for these functions.

This fixes https://github.com/IntelligentSoftwareSystems/Galois/issues/356 and starts work on https://github.com/IntelligentSoftwareSystems/Galois/issues/246 (by deleting an unused header), but quite a bit more work will be needed before that one's fully done.


bozhiyou:
Seems like it's already merged into the Katana repo but not here. Please merge.

insertinterestingnamehere:
It'd be nice to include what they've done there since there were additional changes beyond what I included here. @ddn0 do we have any kind of plan for syncing the two repos?

l-hoang:
from what i know, there really isn't any plan for that that exists at the moment because the code bases have diverged quite significantly, specifically when it comes to the data structures. Donald can elaborate more on it.

fwiw a sync would be very nice to get the gluon/cusp changes in katana that make the code way easier to deal with/extend, but that would likely require some manual sync'ing

PR #363, opened 2020-10-30
Fix broken barrier in DistStats

l-hoang:
Shamelessly stolen from KatanaGraph.

Fixes an issue where host 0 can stay in termination detection because it
detects work after other hosts have decided that they are done with
termination detection.


roshandathathri (approved):
Thanks!

bozhiyou (approved):
LGTM

PR #365, opened 2021-02-18
fix a bug that graph-convert does not work for graph with less than 5…

bigwater:
When I tried to use graph-convert to convert a small graph from mtx to gr format, the graph-convert tool does not work. The error message was `Floating point exception (core dumped)`. 

The bug was in graph-convert.cpp:689. This commit is an easy fix for this problem.


roshandathathri (approved):
LGTM. Thanks!

PR #374, opened 2021-04-23
Update README.md

### LINKED ISSUE #372: CUB not compatible with CUDA version >= 11.0
CUDA 11+ has a builtin CUB module, so the external git module will no longer be necessary and the compiler will complain. Once this is fixed and compiling again, Galois GPU programs will fail with an "illegal memory access," because the version of CUB included in CUDA 11+ is not compatible with Galois.
A note about the CUDA version in the README might be nice for people trying to build for the first time.

MrZmann:
Added more specific CUDA compatibility information to README. 
Fixes #372


MrZmann:
resolves #372

roshandathathri (approved):
Thanks!

PR #377, opened 2021-06-14
Fix 0-initialization of elements in a multiple_sum structure on GPU

srogatch:
Dist apps such as PageRank may crash on some machines because `multiple_sum` structure elements will contain garbage when default-constructed.


roshandathathri (approved):
LGTM

amberhassaan (approved):
No comments

PR #385, opened 2022-02-09
Add functionality to handle self edges in shared memory triangle counting app

uditagar:
This PR adds the functionality to handle self edges in shared memory triangle counting app.


roshandathathri (approved):
Thanks!

PR #413, opened 2024-03-19
fix size of vertex metadata struct

meyerzinn:
Apparently, `galois::SpinLock` is a massive class that takes >128 bytes to represent. This replaces the use of `SpinLock` with a simple atomic byte. In theory, we could combine the `buffer` field and the lock into one flags field, but we already have spare bits in the `begin` and `end` fields so that wouldn't change the size.


meyerzinn:
sorry, wrong repo. ignore.

